PR #101, opened 2017-01-18
Add a Spinner component

tlconnor:
This PR adds a pure CSS spinner component designed to be consistent with the spinner defined in Saffron. https://qa.qa.appfolio.com/saffron/components/loading_indicators#p__example

For a side by side comparison see https://jsfiddle.net/oz14hn08/


gthomas-appfolio:
LGTM

PR #529, opened 2019-03-13
jm - wrap sortabletable with responsiveness

jmangel:


JeremyRH:
No longer supporting scss in react-gears.

PR #713, opened 2020-06-12
Convert ExpandableSection to TypeScript 

### LINKED ISSUE #710: Convert ExpandableSection to TypeScript

YJ83Lee:
Closes #710.

### DIFF
--- a/src/components/ExpandableSection.tsx
+++ b/src/components/ExpandableSection.tsx
@@ -1,22 +1,35 @@
-import PropTypes from 'prop-types';
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, FunctionComponent } from 'react';
 import Collapse from './Collapse';
 import ClickableContainer from './ClickableContainer';
 import Icon from './Icon';
 
-const ExpandableSection = (props) => {
-  const [open, setOpen] = useState(props.open);
+interface ExpandableSectionProps {
+  children?: React.ReactNode;
+  className?: string;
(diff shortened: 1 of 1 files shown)


src/components/ExpandableSection.tsx
```diff
@@ -1,22 +1,35 @@
...
 import Collapse from './Collapse';
 import ClickableContainer from './ClickableContainer';
 import Icon from './Icon';
 
-const ExpandableSection = (props) => {
-  const [open, setOpen] = useState(props.open);
+interface ExpandableSectionProps {
+  children?: React.ReactNode;
+  className?: string;
+  onToggle: (open: boolean) => void;
```
rmacklin, line 9: `ExpandableSection.defaultProps` declares a default `onToggle` prop:
https://github.com/appfolio/react-gears/blob/1adbf65196bce045bc1596a602db6a4e3998a849/src/components/ExpandableSection.tsx#L54-L57
and there are several tests in `ExpandableSection.spec.js` that do not pass the `onToggle` prop:
https://github.com/appfolio/react-gears/blob/1adbf65196bce045bc1596a602db6a4e3998a849/test/components/ExpandableSection.spec.js#L7-L61
so it seems that this `ExpandableSectionProps` interface should define `onToggle` as an optional prop.

rmacklin: This is fixed in https://github.com/appfolio/react-gears/pull/814

PR #730, opened 2020-06-15
bp: convert CurrencyInput to tsx

brucepang:


src/components/CurrencyInput.tsx
```diff
@@ -1,36 +1,62 @@
...
+type Props = {
+  allowDecimal?: boolean;
+  allowNegative?: boolean;
+  className?: string;
+  includeThousandsSeparator?: boolean;
+  inputProps?: { className?: string };
+  size?: string;
+  state?: any;
+  type?: any;
+};
```
rmacklin, line 40: The original type declaration for these props included `React.InputHTMLAttributes<HTMLInputElement>`:
https://github.com/appfolio/react-gears/blob/039ca6a4f058e91cd0e3dd2a40d170224c879120/src/components/CurrencyInput.d.ts#L4
which allowed props like `value`, `onChange`, `onBlur`, etc. to be passed, but this new type definition doesn't allow that.

rmacklin: This is fixed in https://github.com/appfolio/react-gears/pull/814

PR #740, opened 2020-06-26
yl - Convert Paginator to TypeScript

### LINKED ISSUE #739: Paginator (TS, FC)

YJ83Lee:
Closes #739 .

### DIFF
--- a/src/components/Paginator/Page.tsx
+++ b/src/components/Paginator/Page.tsx
@@ -1,31 +1,38 @@
-// @ts-nocheck
 import React from 'react';
 import PropTypes from 'prop-types';
 import PaginationItem from '../PaginationItem';
 import PaginationLink from '../PaginationLink';
 
+type PageProps = {
+  current: boolean,
+  onClick: (pageNumber: number) => void,
+  page: number,
+}
+
 /**
  * A clickable link to a page in the pagination bar
  */
-export default class Page extends React.Component {
-  static propTypes = {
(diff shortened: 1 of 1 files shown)


src/components/Paginator.tsx
```diff
@@ -0,0 +1,133 @@
...
+PrevPageLink.propTypes = linkPropTypes;
+NextPageLink.propTypes = linkPropTypes;
+LastPageLink.propTypes = linkPropTypes;
+
+type PaginatorProps = {
+  currentPage: number,
+  onClick: (page: number) => void,
+  perPage?: number,
+  size?: 'sm' | 'lg',
+  summary: React.ReactNode,
```
rmacklin, line 59: The `Paginator` component has built-in handling for the absence of the `summary` prop:
https://github.com/appfolio/react-gears/blob/1adbf65196bce045bc1596a602db6a4e3998a849/src/components/Paginator.tsx#L119
and there are numerous tests in `Paginator.spec.js` that do not pass the `summary` prop:
https://github.com/mycase/react-gears/blob/1adbf65196bce045bc1596a602db6a4e3998a849/test/components/Paginator.spec.js#L8-L189
so it seems that the `PaginatorProps` type should define `summary` as an optional prop.

rmacklin: This is fixed in https://github.com/appfolio/react-gears/pull/814

PR #750, opened 2020-07-09
create story showing bug with using uncontrolled inputs in HMF …

tyleralves:
…templates


JeremyRH:
HasManyFields has been refactored significantly.

PR #762, opened 2020-08-03
Major version update to date-fns

joshuasbates:
During Hack Day, I was planning to put some code into `react-gears` that required `date-fns-tz` to perform timezone operations, and this module requires `date-fns` 2.0.0 or greater.  As a result, I spent some time to see what it would take to perform that update, and the result is this PR.

Unfortunately, because `react-gears` exposes the date format in a couple places, and the new version of `date-fns` deprecates some of the previous formatting rules in favor of UNICODE standards (for example, "year" changed from `Y` to `y` syntax, day from `D` to `d` and AM/PM from `A` to `a`), `react-gears` clients that specify date formats for affected components will be impacted.

In the end, I found a better home for my time-zone-respecting components, so this update is not required, but it still may be desired to keep `react-gears` up-to-date.

As such, I present this PR as a draft to demonstrate the changes that will be required to update to `date-fns` 2.15.0.  I leave it to the `react-gears` dev team to do with this what they will.  😄

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11400,9 +11400,9 @@
       }
     },
     "date-fns": {
-      "version": "1.30.1",
-      "resolved": "https://registry.npmjs.org/date-fns/-/date-fns-1.30.1.tgz",
-      "integrity": "sha512-hBSVCvSmWC+QypYObzwGOd9wqdDpOt+0wl0KbU+R+uuZBS1jN8VsD1ss3irQDknRj5NvxiTF6oj/nDRnN/UQNw=="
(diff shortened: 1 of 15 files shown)


JeremyRH:
Another attempt to update date-fns here: https://github.com/appfolio/react-gears/pull/1029/files

Unfortunately, the format string is a prop for some components meaning we don't have control of all the inputs to data-fns. This is a major breaking change for these components. It's easier to just re-write them in a new library and convert usage of these components to the new ones.

PR #807, opened 2020-12-14
allow passing id to MonthInput

darreneng:
needed for a11y


kkirasich (approved):
Really great work :highfive:

PR #812, opened 2021-01-05
feat: pass thru className of FormChoice elements in a CheckboxListInput

tkissingappfolio:
This allows among other things to indent a checkbox (including label) in a single checkbox list.

### DIFF
--- a/src/components/FormChoice.js
+++ b/src/components/FormChoice.js
@@ -1,6 +1,7 @@
-import React from 'react';
-import PropTypes from 'prop-types';
 import classname from 'classnames';
+import PropTypes from 'prop-types';
+import React from 'react';
+
 import FormGroup from './FormGroup';
 import Input from './Input';
 import Label from './Label';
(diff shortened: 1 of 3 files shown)


kevinkir (approved):

src/components/FormChoice.js
```diff
@@ -45,12 +48,12 @@ class FormChoice extends React.Component {
       );
     }
 
-    const labelClasses = classname({ 'form-check-inline': inline });
+    const containerClasses = classname({ 'form-check-inline': inline }, className);
```
gthomas-appfolio, line 51: This change makes sense, however I do see some usages (likely for tests) that pass className now and end up in the `<Input {...attributes}>` below.  E.g. https://github.com/appfolio/apm_bundle/blob/0e484c00accb49211be3fcd3ecd2e43a088e9e45/apps/property/app/javascript/move_outs/src/components/accounting/splitSecurityDeposit/splitRefundRadios.js#L25


Could we consider calling this `containerClass` etc or could we get your help updating the usages to avoid a major version bump for this?

gthomas-appfolio (commented):
One question

src/components/FormChoice.js
tkissingappfolio: Do you know how to get to that page? From my testing yesterday that `className='js-split-refund-radio-no'` will not show up anywhere in the actual HTML.

tkissingappfolio: Just to be clear: I am not opposed to renaming, I just want to make sure I didn't miss a code path that passes a className thru.

gthomas-appfolio (approved):

gthomas-appfolio (approved):

PR #818, opened 2021-01-26
Update <FeatureBanner /> subtitle type

YJ83Lee:
With the previous type, if a user wants the subtitle to have a link
and passed in a react component.

E.g
```jsx
const subtitle = <span>More info in <a href={url}here</a>.</span>
<FeatureBanner
  alertText="New"
  subtitle={subtitle}
  title="Title"
/>
```

The user gets prop type warning or TypeScript error while it works
perfectly.

The type of subtitle prop of <FeatureBanner /> was string, while
it can be anything that can be rendered, which is PropTypes.node,
React.ReactNode.
https://github.com/appfolio/react-gears/blob/c0e0eebbead311527be3b8b119fc94c3dc7f3a3d/src/components/FeatureBanner.js#L35

### DIFF
--- a/src/components/FeatureBanner.d.ts
+++ b/src/components/FeatureBanner.d.ts
@@ -3,7 +3,7 @@ interface FeatureBannerProps {
   alertText?: string;
   children?: React.ReactNode;
   color?: string;
-  subtitle: string;
+  subtitle: React.ReactNode;
   title: string;
 }
 declare class FeatureBanner extends React.Component<FeatureBannerProps, {}> { }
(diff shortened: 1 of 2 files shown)


YJ83Lee:
With https://github.com/appfolio/react-gears/blob/c0e0eebbead311527be3b8b119fc94c3dc7f3a3d/src/components/FeatureBanner.js#L35
it is true that not all html element should be inside `<p>` element (e.g., we should not have a `<div>` inside `<p>`).
But I think it is still good to loosen the type to not be `string` so that having `<a>` inside no longer causes prop type warning or TypeScript error.

gthomas-appfolio (approved):

Issue #834, opened 2021-04-01
Request for mapping address values with custom names


joyjoy993:
Hi, I realize that `AddressInput` has a prop `inputName` that allows us to map custom field names. It would be nice that we can map address values with custom field names as well.
In IM, our address model looks like this:
```
city: "sfdsx"
countryCode: "US"
postalCode: "11220"
state: "AR"
street1: "xxxx"
street2: ""
```
so every time when we pass or fetch the value from `AddressInput`, we need to do a conversion.
I'm wondering if we can have another prop something like `mapValuesWithCustomFields` to convert values here https://github.com/appfolio/react-gears/blob/master/src/components/AddressInput.js#L68
like
```
propsFor = (field) => {
	const value = mapValuesWithCustomFields ? this.props.value[this.nameFor(field)] : this.props.value[field];
	const defaultValue = mapValuesWithCustomFields ? this.props.defaultValue[this.nameFor(field)] : this.props.defaultValue[field];
	if (value !== undefined) {
	  return { value };
	}
	return { defaultValue };
}
```
and then we can pass
```
<AddressInput
	inputName={{
		address1: 'street1',
		address2: 'street2',
		state: 'state',
		postal: 'postalCode',
		countryCode: 'countryCode',
		city: 'city'
	}}
	mapValuesWithCustomFields={true}
	......
```
This would help us to simplify the logic a lot, I don't know what's the cost of this change from RG standpoint, if it's feasible, I'm happy to drop a PR to do this.
Thanks.

phwebi:
@joyjoy993 I looked into this and it actually seems like a bug from the original feature to add custom names to the fields! I think we should automatically use the custom names if they're provided. Seems like an easy enough fix, so happy to tackle that this morning and pass it over for you to review if you don't mind. (Just to make sure it'll work for your use case)

PR #835, opened 2021-04-02
Make AddressInput respect custom input names everywhere

phwebi:
This PR updates `AddressInput` such that if custom input names are used, they will be respected everywhere.

Breaking changes:

- `value` and `defaultValue` props will always use input names as keys within the value object. This means if you use custom names, your value props will need to match the custom names.  (Addresses issue: https://github.com/appfolio/react-gears/issues/834)
- `onBlur` will now be passed the actual input name instead of just the default name.

### DIFF
--- a/src/components/AddressInput.js
+++ b/src/components/AddressInput.js
@@ -187,7 +187,7 @@ class AddressInput extends React.Component {
                 {...this.propsFor(inputName.state)}
                 invalid={!!error.state}
                 onBlur={() => onBlur(inputName.state)}
-                onChange={state => this.onChange({ state })}
+                onChange={state => this.onChange({ [inputName.state]: state })}
                 disabled={disabled}
               />
             </FormLabelGroup>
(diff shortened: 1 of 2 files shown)


src/components/AddressInput.js
```diff
@@ -177,11 +182,11 @@ class AddressInput extends React.Component {
...
                 id={stateId}
-                name={this.nameFor('state')}
+                name={inputName.state}
                 placeholder={labels.state}
-                {...this.propsFor('state')}
+                {...this.propsFor(inputName.state)}
                 invalid={!!error.state}
-                onBlur={() => onBlur('state')}
+                onBlur={() => onBlur(inputName.state)}
                 onChange={state => this.onChange({ state })}
```
joyjoy993, line 190: I think the `onChange` should change based on the custom names as well, it's not working now...

joyjoy993 (changes requested):
Thanks for doing this! just another thing, do you think it's worth a test in `test/components/AddressInput.spec.js`? not sure what the test plan of RG 😅

src/components/AddressInput.js
phwebi: Ooh yes - didn't notice that that there were two that weren't using the flow thing.

joyjoy993 (approved):
Thank you! 👍

PR #836, opened 2021-04-07
support localized date formatting in DateInput

phwebi:
This PR adds localization support for `DateInput`. It introduces a new `locale` prop that supports `date-fns` locale objects that will be used to format dates in the component.

### DIFF
--- a/src/components/Calendar.js
+++ b/src/components/Calendar.js
@@ -16,8 +16,6 @@ import startOfWeek from 'date-fns/start_of_week';
 import enLocale from 'date-fns/locale/en';
 import Table from './Table';
 
-// TODO locale/localize
-
 const Day = ({ day, dateFormat, locale, onClick, ...props }) => {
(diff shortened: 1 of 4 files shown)


src/components/Calendar.js
```diff
@@ -13,11 +13,12 @@ import isToday from 'date-fns/is_today';
 import startOfDay from 'date-fns/start_of_day';
 import startOfMonth from 'date-fns/start_of_month';
 import startOfWeek from 'date-fns/start_of_week';
+import enLocale from 'date-fns/locale/en';
 import Table from './Table';
 
 // TODO locale/localize
```
ltewksbury, line 19: I feel like you can delete this comment now.

ltewksbury (approved):

mathewbaltes (approved):
One comment is to refactor the explicit classnames to use another approach for testing, but looks like this is an existing pattern.  Other than that looks good.

PR #838, opened 2021-04-14
Replace react text mask

phwebi:
This PR migrates CurrencyInput and MaskedInput to use react-imask instead of react-text-mask (which is no longer being maintained).

#### CurrencyInput Updates
- new `padZeros` property that defaults to true

#### MaskedInput Updates
- MaskedInput is now a wrapper for react-imask instead of react-text-mask. Props and usage are different. See https://imask.js.org/ for docs.


src/components/CurrencyInput.tsx
```diff
@@ -38,12 +38,20 @@ const CurrencyInput: FunctionComponent<Props> = ({
...
   value,
-  onChange: onAccept,
+  onChange,
   ...props
 }: Props) => {
   const inputClassNames = classnames('form-control', inputProps && inputProps.className);
+  const onAccept = (val: string, mask: IMask.InputMask<IMask.MaskedNumberOptions>) => {
+    if (onChange) {
+      const ev = new Event('change') as unknown as React.ChangeEvent<HTMLInputElement>;
+      // @ts-ignore
```
phwebi, line 48: Needed a couple ts-ignore comments because the imask typings seem to have some mistakes/stuff missing. Can look into fixing it later.

gthomas-appfolio (approved):

phwebi:
Going to split this up into two PRs, so we won't need a major version update for the CurrencyInput changes.

PR #840, opened 2021-04-19
fix "undo" style in delete note alert; change example className to ma…

Brick-Mover:
…tch apm style

### DIFF
--- a/src/components/DeletedNote.tsx
+++ b/src/components/DeletedNote.tsx
@@ -20,7 +20,7 @@ const DeletedNote: React.FunctionComponent<DeletedNoteProps> = ({
       <span className="align-middle">Note deleted.</span>
       {onUndelete &&
         <Button color="link" onClick={() => onUndelete(note)} className="ml-1 p-0">
-          undo
+          Undo
         </Button>
       }
     </Alert>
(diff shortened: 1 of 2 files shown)


Brick-Mover:
[Change flash message color and undo UX for notes block](https://trello.com/c/Q3kNCXax/1499-change-flash-message-color-and-undo-ux-for-notes-block)

MMullin7 (approved):

src/components/DeletedNote.tsx
```diff
@@ -20,7 +20,7 @@ const DeletedNote: React.FunctionComponent<DeletedNoteProps> = ({
       <span className="align-middle">Note deleted.</span>
       {onUndelete &&
         <Button color="link" onClick={() => onUndelete(note)} className="ml-1 p-0">
-          undo
+          <u>Undo</u>
```
gthomas-appfolio, line 23: Why is this a u tag?  Seems odd

gthomas-appfolio (commented):
One question

src/components/DeletedNote.tsx
Brick-Mover: Yeah so we want to add underscore to "Undo" button to make it more obvious

PR #844, opened 2021-04-27
Change Select border styling to match Coastline input styling

tyleralves:
**Note:** I did not attempt to incorporate the Coastline stylesheet in any way. This is a quick fix to update the styling closer to current Coastline styling.

.Select-control: We remove `border-color` since it is already applied in
the `border` property.

.Select.is-open > .Select-control: We remove `border-color` to match the
normal Input styling, which doesn't change when it is open.

Updated `Select`:
image

Existing `Input type='select':
image


tyleralves:
Closing in favor of a Coastline change

PR #845, opened 2021-04-30
Update to TS 4

gthomas-appfolio:
Remove the dupe PropTypes in TS components


phwebi:
Update looks good, but I think we left the js proptypes so support prop type warnings for non TS apps. Not sure if we still want to keep those.

PR #848, opened 2021-05-03
Add create-able options for Combobox

phwebi:
Add create-able options for Combobox.

The creatable options of combobox differ from the `Select` in that options are completely controlled. Instead of having `creatable` and `onNewOptionClick` props. We simplified it to an `onCreate` prop that's required if you want to enable creatable options.

`onCreate` should return the value for the newly created option or a falsey value if creation failed (i.e. failed validation).
`isValidNewOption` takes in a string (input value) and returns a boolean

### DIFF
--- a/src/components/Combobox.tsx
+++ b/src/components/Combobox.tsx
@@ -231,7 +231,7 @@ function Combobox<T>({
             createOption();
           }}
         >
-          {`Add option: ${inputValue}`}
+          {`Create "${inputValue}"`}
         </DropdownItem>
       );
     }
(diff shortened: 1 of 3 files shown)


src/components/Combobox.tsx
```diff
@@ -149,11 +152,23 @@ function Combobox<T>({
     }
   };
 
+  const createOption = () => {
+    if (!onCreate) return;
+
+    const optionValue = onCreate(inputValue);
+    selectOption(optionValue);
```
phwebi, line 159: It seemed to make sense to select the option on creation - interested to hear thoughts on this.

gthomas-appfolio (approved):
Supercalifragilisticexpialidocius

PR #851, opened 2021-05-07
Remove modal overrides

zldavis:


aaronmars:
This PR is way out of date. Will revisit if necessary

PR #852, opened 2021-05-07
Fix accessibility issues with HelpBubble

darreneng:
Right now, the help bubble icon isn't a button and doesn't have an accessible name, so some users may not know / aren't able to click it.

Even if screenreader users can do so, they wouldn't know what the description text is because it appears in a different location in the DOM.


src/components/HelpBubble.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const toggle = (e: SyntheticEvent) => {
+    e.stopPropagation();
+    setIsOpen((prevIsOpen) => !prevIsOpen);
+  };
+
+  const { title, children, className, ...other } = props;
+
+  return (
+    <span className={className} style={style}>
+      <ClickableContainer aria-label="More Info" tag="span" onClick={toggle}>
```
gthomas-appfolio, line 44: Would it be better to use `<Button color="link">` with the appropriate padding?  ClickableContainer was kind of a workaround for bigger areas like headers, etc.

gthomas-appfolio (commented):
One question, thanks for the a11y fixes

src/components/HelpBubble.tsx
darreneng: yeah I agree, I would much rather use an actual `button` tag. 

`<Button color="link">` seems to mostly work, but we lose the outline style applied when the help bubble receives focus - not sure how to get that to show up w/o custom styles?


Screen Shot 2021-05-10 at 5 26 30 PM
Screen Shot 2021-05-10 at 5 28 14 PM

gthomas-appfolio: Hmm, let me see what the delta issue

gthomas-appfolio: Looks like it's bootstrap removing box-shadow on :focus in favor of text underline:

Screen Shot 2021-05-11 at 1 41 03 PM

Perhaps we should use `<Button color="link">` and adjust focus styles in Bootstrap theme?  Given the icon does not show underline

PR #853, opened 2021-05-11
Example - Combobox async options support

zldavis:
As it currently, it does not seem possible to use the `<Combobox />` for async/search like functionality. This is just an example and does not fully work, but wondering what ya'll think about adding support for something like this? I would love to not have to keep using the `<Select />` component for multi-async search 😃.


aaronmars:
Looking to replace `Combobox`/`Select` with a proper wrapper of `react-select`.
This PR is no longer necessary.

PR #861, opened 2021-05-13
feat: make steps clickable

kevinkir:
This adds an optional prop called "onStepClick" to the Steps component.
If it is provided, then it it will be called with the step index when
a step is clicked.

### DIFF
--- a/stories/Steps.stories.js
+++ b/stories/Steps.stories.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { action } from '@storybook/addon-actions';
 import { boolean, number } from '@storybook/addon-knobs';
 import { Steps } from '../src';
 
@@ -31,5 +32,17 @@ export const Vertical = () => (
     complete={boolean('complete', false)}
     vertical={boolean('vertical', true)}
     collapse={boolean('collapse', false)}
+  />
+);
+
+export const Clickable = () => (
+  <Steps
+    steps={steps}
+    step={number('step', 0)}
+    complete={boolean('complete', false)}
+    vertical={boolean('vertical', false)}
+    collapse={boolean('collapse', false)}
+    // eslint-disable-next-line no-alert
+    onStepClick={action('index')}
   />
 );
--- a/src/components/Steps.tsx
+++ b/src/components/Steps.tsx
@@ -1,16 +1,18 @@
 import React from 'react';
 import classNames from 'classnames';
 import Icon from './Icon';
+import Button from './Button';
 
 interface StepProps {
   collapse?: boolean;
   complete?: boolean;
+  onStepClick?: (stepIndex: number) => void;
   step?: number;
   steps?: string[];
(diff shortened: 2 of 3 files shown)


kevinkir:
[Business hours setup UI](https://trello.com/c/r4ONUoYx/682-business-hours-setup-ui)

tkissingappfolio (approved):
LGTM

gthomas-appfolio:
Thanks.  @kevinkir ,  I'm wondering if it would be preferable to allow React nodes as steps, then this interface would be allowed:

```javascript
const steps = [
  <a onClick={whatever}>Put your left foot in'</a>,
  <a onClick={whatever}>Put your right foot out'</a>,
  <a onClick={whatever}>Do the Hokey-Pokey'</a>,
  <a onClick={whatever}>Turn yourself about</a>
];

<Steps steps={steps}  step={0} />
```

gthomas-appfolio (commented):
Thanks, could you take a look at idea in comments.

kevinkir:
> Thanks. @kevinkir , I'm wondering if it would be preferable to allow React nodes as steps, then this interface would be allowed:
> 
> ```js
> const steps = [
>   <a onClick={whatever}>Put your left foot in'</a>,
>   <a onClick={whatever}>Put your right foot out'</a>,
>   <a onClick={whatever}>Do the Hokey-Pokey'</a>,
>   <a onClick={whatever}>Turn yourself about</a>
> ];
> 
> <Steps steps={steps}  step={0} />
> ```

I like that idea, since it would be nice to be able to provide a react-router Link, and it might improve accessibility. My concern, though, would be that this with this implementation, only the text would be clickable and not the bubbles with the numbers in them, which I don't think is a very intuitive user experience.

stories/Steps.stories.js
```diff
@@ -33,3 +33,15 @@ export const Vertical = () => (
...
+
+export const Clickable = () => (
+  <Steps
+    steps={steps}
+    step={number('step', 0)}
+    complete={boolean('complete', false)}
+    vertical={boolean('vertical', false)}
+    collapse={boolean('collapse', false)}
+    // eslint-disable-next-line no-alert
+    onStepClick={i => alert(`You clicked step ${i}!`)}
```
phwebi, line 45: Instead of doing an alert, we can use storybook actions and storybook will continuously log the function calls/params directly in the right side panel! (See [example](https://github.com/appfolio/react-gears/blob/master/stories/Table.stories.js#L110))

image

phwebi (commented):
Looks good! Just one comment on the `Steps` story.

stories/Steps.stories.js
kevinkir: Ah, neat! Figured there must be a better way. Switched it to use `action('index')`.

phwebi: 🎉

phwebi (approved):

PR #864, opened 2021-05-14
Add onInputChange to support async loading options

phwebi:


aaronmars:
Combobox will be replaced by a fresh wrapper of `react-select`. This use case will be handled at that time.

PR #870, opened 2021-06-22
Combobox: add support for object values

phwebi:


src/components/Combobox.tsx
```diff
@@ -193,8 +194,8 @@ function Combobox<T>({
       <DropdownItem
         disabled={option.disabled}
         className={`${isOptionVisible(option) ? '' : 'sr-only'}`}
-        key={`${option.value}`}
-        id={`option-${option.value}`}
+        key={`${JSON.stringify(option.value)}`}
```
mathewbaltes, line 197: We should only JSON.stringify() once.  Doing this multiple times will be expensive.

src/components/Combobox.tsx
```diff
@@ -78,7 +79,7 @@ function Combobox<T>({
     }
     return optionsProp as Option<T>[];
   }, [optionsProp, grouped]);
-  const selected = useMemo<Option<T> | Option<T>[]>(() => multi ? (value || []).map((v: T) => options.find(option => option.value === v)) : options.find(option => option.value === value), [value, options, multi]);
+  const selected = useMemo<Option<T> | Option<T>[]>(() => multi ? (value || []).map((v: T) => options.find(option => equal(option.value, v))) : options.find(option => equal(option.value, value)), [value, options, multi]);
```
mathewbaltes, line 82: Do we need this? Seems like the order of the JSON.stringify() should be equivalent each time.

mathewbaltes: If we do use JSON.stringify() here, we could reuse as well for the dropdown.

mathewbaltes: Also, using equal each time is something we should be able to cache when we first pass the options in.  We shouldn't do this each time we have a selection.

phwebi: I feel better actually doing a deep comparison of the object because we don't know what might happen in the external `onChange` handlers. It is possible for the order of the object keys in the `value` prop to be different than the option value.

phwebi: > Also, using equal each time is something we should be able to cache when we first pass the options in. We shouldn't do this each time we have a selection.

`useMemo` should actually take care of this. It's only recomputed when dependencies change.

mathewbaltes (approved):

PR #873, opened 2021-07-07
Add 'isValidElement' check to FormRow.

kzhang3702:
The reason for this change is that we want to support conditional rendering
of children. Eg.
```
<FormRow>
  <FormChoice />
  {conditional && <FormChoice />}
<FormRow>
```
should render appropriately.

### DIFF
--- a/src/components/FormRow.js
+++ b/src/components/FormRow.js
@@ -96,7 +96,7 @@ const FormRow = (props) => {
         {...childFeedback}
       >
         {shouldPassChildren ? React.Children.map(children, child =>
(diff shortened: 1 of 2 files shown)


kzhang3702:
[Add Adjustable-Rate Loan to the Loan Type dropdown](https://trello.com/c/LjRGeO3c/1437-add-adjustable-rate-loan-to-the-loan-type-dropdown)

almgong (commented):
Changes make sense to me, but the lint errors are need to be addressed.

almgong (approved):

PR #882, opened 2021-08-25
UncontrolledTable - add support for server side pagination

phwebi:
- new `loadRows` prop to handle async loading of rows based on pagination/sorting params
- added placeholder rows for loading state


aaronmars:
Table is getting a complete refresh. This use case will be handled at that time.

PR #884, opened 2021-09-07
Support default `type` prop for `Status` component

darreneng:
The `Status` component already implicitly supports not specifying the type prop, e.g. `<Status />`.

In typescript, however, this causes an error because the `type` prop is not marked as optional.

This PR supports omitting the prop value in TS as well as an explicit value for this state.

### DIFF
--- a/src/components/Status.tsx
+++ b/src/components/Status.tsx
@@ -4,12 +4,12 @@ import Icon from './Icon';
 import { FontAwesomeAPMProps } from './icon/FontAwesomeAPM';
 
 interface StatusProps extends Omit<FontAwesomeAPMProps, 'name'> {
-  type: 'info' | 'muted' | 'success' | 'danger' | 'warning';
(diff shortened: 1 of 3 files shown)


src/components/Status.tsx
```diff
@@ -4,12 +4,12 @@ import Icon from './Icon';
...
 
 interface StatusProps extends Omit<FontAwesomeAPMProps, 'name'> {
-  type: 'info' | 'muted' | 'success' | 'danger' | 'warning';
+  type?: 'info' | 'muted' | 'success' | 'danger' | 'warning' | 'none';
   className?: string;
 }
 
-const Status = ({ type, className, ...props }: StatusProps) => {
-  let name = 'circle';
+const Status = ({ type = 'none', className, ...props }: StatusProps) => {
```
darreneng, line 11: not really picky on the name `none` - open to changing this to something else

phwebi (approved):

PR #890, opened 2021-09-15
Add prop to control col widths

zldavis:
Since the column widths are based on the viewport size, if the `Address` component is rendered within an element with a smaller width, the city, state, and postal inputs can get squished and unreadable in some cases.
This makes it so you can override those column widths to handle instances where this could be a problem.

Example before:
Screen Shot 2021-09-15 at 6 36 50 PM

After with custom col widths pushing the city to `xs: 12`:
Screen Shot 2021-09-15 at 6 36 17 PM

### DIFF
--- a/src/components/AddressInput.d.ts
+++ b/src/components/AddressInput.d.ts
@@ -11,6 +11,14 @@ interface AddressInterface {
 
 type AddressNames = keyof AddressInterface;
 
+interface ColWidthInterface {
+  xs?: number;
+  sm?: number;
+  md?: number;
+  lg?: number;
+  xl?: number;
+}
+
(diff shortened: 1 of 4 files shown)


gthomas-appfolio (commented):
I'm wondering now if using rows+cols for this component is not the right approach to begin with.  Your use case makes sense, and basically the state and zipcodes are fixed width, and the city fills remaining space (with some min width itself).

How about I'll take a look at a flex or grid approach that could handle this better before we merge this.

zldavis:
sure sounds good, thanks!

phwebi:
Closing since we decided to go with adding a compact mode to `AddressInput` (#895)

PR #895, opened 2021-09-28
Add compact mode to AddressInput

phwebi:
This should solve the root problem of https://github.com/appfolio/react-gears/pull/890

### DIFF
--- a/src/components/AddressInput.js
+++ b/src/components/AddressInput.js
@@ -24,6 +24,7 @@ export const addressPropType = {
 class AddressInput extends React.Component {
   static propTypes = {
     className: PropTypes.string,
+    compact: PropTypes.bool,
     countries: PropTypes.arrayOf(PropTypes.string),
     defaultValue: PropTypes.object,
     inputName: PropTypes.shape(addressPropType),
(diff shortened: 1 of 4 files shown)


fabio-aiello-appfolio (approved):

src/components/AddressInput.js
```diff
@@ -143,7 +144,7 @@ class AddressInput extends React.Component {
           />
         </FormLabelGroup>
         <Row className="no-gutters">
-          <Col sm={6} xs={12} className="pr-sm-3">
+          <Col sm={compact ? undefined : 6} xs={12} className="pr-sm-3">
```
zldavis, line 147: I think we should get rid of the `pr` class when in compact mode similar to the col size logic:

```
<Col sm={compact ? undefined : 6} xs={12} className={compact ? undefined: 'pr-sm-3'}>
```

Otherwise can lead to some funky padding when the viewport size is >= `sm`

Screen Shot 2021-09-28 at 5 09 26 PM

emcpete (commented):
Looks good

zldavis:
Also some issues with the bottom margin when `showCountry={false}`

Screen Shot 2021-09-28 at 5 33 06 PM

PR #916, opened 2021-10-19
add a conditional prop that controls if a row is selectable

Brick-Mover:


Brick-Mover:
[Sortable List Checkbox Disable Spike](https://trello.com/c/ikz05P2E/1846-sortable-list-checkbox-disable-spike)

PR #931, opened 2021-11-23
Imrprove FormRow inner component's props typing

steventhan:


src/components/FormRow.d.ts
```diff
@@ -1,27 +1,62 @@
...
+};
+
+type FormRowProps<TComponent extends NestedComponent> = Omit<
+  ComponentProps<Lookup<TComponent>>,
+  keyof BaseFormRowProps<TComponent>
+> &
+  BaseFormRowProps<TComponent>;
+
+declare const FormRow: <TComponent extends NestedComponent = 'input'>(
+  props: PropsWithChildren<FormRowProps<TComponent>>
```
davidacevedo, line 59: `PropsWithChildren` is not needed as you've already defined children in the props above.

steventhan: fixed in 23ad41d, decided to remove `children` from `BaseFormRowProps` and use `PropsWithChildren` to keep our types consistent with `react`

davidacevedo (approved):

PR #936, opened 2021-12-23
Add allowed size types for font awesome icons

harrypresman:
Only adds the smallest 2 sizes.  There are still a bunch of larger
sizes (6x-10x) allowed by FA that we may also need to add in the future.


davidacevedo (approved):
Looks good 👍

PR #940, opened 2021-12-30
Add accessible text to Status

darreneng:
When rendering a status, we should always provide accessible text
because the icon is hidden to a screenreader via the `aria-hidden`
attribute. We also provide the `title` attribute for sighted users when
hovering with a mouse.

https://fontawesome.com/v5.15/how-to-use/on-the-web/other-topics/accessibility#web-fonts-semantic


aaronmars:
@darreneng: I know this is a very old issue, but would it be possible to take another look at it and perhaps see about using the `ariaLabel` prop on the Icon? That should do what we want.

Also, many of the awesome testing improvements you included here are now part of `react-gears`.

aaronmars:
Closing since `ariaLabel` prop is the preferred method of achieving this, and the testing improvements have been implemented elsewhere.

PR #941, opened 2022-01-07
Add component for truncating text

daviditsygin:
A component that can truncate text to a character limit. It can optionally display the original text in a `Tooltip` on hover.

image

### DIFF
--- a/src/components/TruncatedText.tsx
+++ b/src/components/TruncatedText.tsx
@@ -1,0 +1,41 @@
+import React, { useState } from 'react';
+import { TooltipProps, UncontrolledTooltipProps } from 'reactstrap';
+import Tooltip from './Tooltip';
+
+
+interface TruncatedTextProps {
+  text: string;
+  targetId: string;
+  charLimit: number;
+  tooltip: boolean;
+  placement: UncontrolledTooltipProps['placement'];
+  tooltipProps: TooltipProps
+}
+
+let count = 0;
+function getID() {
+  return `truncated-text-${count++}`; // eslint-disable-line no-plusplus
+}
+const TruncatedText = ({ text, targetId, charLimit = 50, tooltip = true, placement = 'top', tooltipProps }: TruncatedTextProps) => {
+  const [id] = useState(targetId || getID());
+  return (
+    <>
+      <span id={id}>
+        {text && text.length > charLimit ? `${text.substring(0, charLimit).trim()}...` : text}
+      </span>
+      {(text && text.length > charLimit && tooltip) && (
+        <Tooltip
+          {...tooltipProps}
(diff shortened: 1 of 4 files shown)


src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { TooltipProps } from 'reactstrap';
+import Tooltip from './Tooltip';
+
+
+interface TruncatedTextProps extends TooltipProps {
+  text: string;
+  targetId: string;
+  recordId: string | number;
```
davidacevedo, line 9: I don't think having both targetId and recordId makes sense in this case. We should only really need a prop called id and this logic for making an id that comprises of two variables should be on the user side instead.

src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,34 @@
...
+  charLimit: number;
+  tooltip: boolean;
```
davidacevedo, line 11: It might be better to include a `tooltipProps` key here. It seems like the top level props is for the truncated text and it so happens to also include tooltipProps.

src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+const TruncatedText = ({ text, targetId, recordId, charLimit = 50, tooltip = true, placement = 'top', ...props }: TruncatedTextProps) => {
+  return (
```
davidacevedo, line 15: For others that are not making their own ids https://github.com/appfolio/react-gears/blob/d50ab4cd0a10295ecba6ebf5413e8cf9d4b39f17/src/components/FormChoice.js#L11-L13.
```suggestion
  const [id] = useState(id || getId())

  return (
```

test/components/TruncatedText.spec.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import assert from 'assert';
+import { mount } from 'enzyme';
```
davidacevedo, line 3: Nothing to do now, but for the next time you contribute, we are starting to prefer react-testing-library over enzyme. So we do prefer that is used instead.

davidacevedo:
Should fix the linting errors

src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { useState } from 'react';
+import { TooltipProps } from 'reactstrap';
+import Tooltip from './Tooltip';
+
+
+interface TruncatedTextProps extends TooltipProps {
```
davidacevedo, line 6: Does not need to extend `TooltipProps` anymore.

davidacevedo (approved):

PR #943, opened 2022-01-20
Add visible rows notification to UncontrolledTable

aaronmars:
- Add new prop, `onVisibleRowsChanged` to notify when a new set of rows is (potentially) loaded in the current page
- Use `componentDidUpdate` to inspect the state/props that might indicate an updated set of rows
- Add a couple of instance methods to help in getting the rows that should be visible, and to determine it things changed
- Tests and story updated

### DIFF
--- a/src/components/UncontrolledTable.js
+++ b/src/components/UncontrolledTable.js
@@ -20,7 +20,8 @@ export default class UncontrolledTable extends React.Component {
       ascending: PropTypes.bool
     }),
     onSelect: PropTypes.func,
-    onSort: PropTypes.func
+    onSort: PropTypes.func,
+    onVisibleRowsChange: PropTypes.func
   };
 
   static defaultProps = {
@@ -35,7 +36,8 @@ export default class UncontrolledTable extends React.Component {
       ascending: true
     },
     onSelect: () => {},
-    onSort: () => {}
+    onSort: () => {},
+    onVisibleRowsChange: () => {}
   };
 
   state = {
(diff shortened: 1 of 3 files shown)


src/components/UncontrolledTable.js
```diff
@@ -20,7 +20,8 @@ export default class UncontrolledTable extends React.Component {
       ascending: PropTypes.bool
     }),
     onSelect: PropTypes.func,
-    onSort: PropTypes.func
+    onSort: PropTypes.func,
+    onVisibleRowsChanged: PropTypes.func
```
steventhan, line 24: I think this should just be `onVisibleRowsChange` to keep it consistent with react conversion, we also have an `onPageChange` prop couple lines up
```suggestion
    onVisibleRowsChange: PropTypes.func
```

steventhan (commented):
Looks great, just a tiny comment

src/components/UncontrolledTable.js
aaronmars: Good catch! Changing.

steventhan (approved):

src/components/UncontrolledTable.js
```diff
@@ -129,6 +131,34 @@ export default class UncontrolledTable extends React.Component {
...
+
+    const { page: prevPage } = prevState;
+    const { ascending: prevAscending, column: prevColumn } = prevState.sort;
+    const { paginated: prevPaginated, pageSize: prevPageSize, rows: prevRows } = prevProps;
+
+    return page !== prevPage ||
+      ascending !== prevAscending ||
+      column !== prevColumn ||
+      paginated !== prevPaginated ||
+      pageSize !== prevPageSize ||
```
davidacevedo, line 158: This is a good example of where converting to a functional component could make this a bit easier, which you don't need to do now!

But for an example, this would be the way with hooks:
```
useEffect(() => {
  onVisibleRowsChange(getVisibleRows());
}, [page, ascending, column, paginated, pageSize, rows])
```
Or something similar to the above.

aaronmars: I 100% agree. Converting to a functional component was my first impulse (followed closely by converting to TS)!

src/components/UncontrolledTable.js
```diff
@@ -129,6 +131,34 @@ export default class UncontrolledTable extends React.Component {
...
+      !this.isEqualUsingKeys(rows, prevRows);
```
davidacevedo, line 159: No actionable thought for this comment, but I wonder if doing `rows !== prevRows` would be enough as I would assume that the object reference would change if someone created a new object with new rows. Although someone can easily just do something like this:
```
<Table rows={{ ...myItems }} />
```
Which will trigger the onChange all the time 😄

davidacevedo (approved):

PR #947, opened 2022-01-24
fix: report visible rows on initial render

aaronmars:
In https://github.com/appfolio/react-gears/pull/943, logic was added to `componenDidUpdate()` to call the `onVisibleRowsChange` prop when the props/state changed in a way that would cause the contents of the current table page to change.

Unfortunately, this did not take in to account the initial render. `componenDidUpdate()` is not called at that time. In this PR, we call `componenDidMount()` to get that initial call to `onVisibleRowsChange`

### DIFF
--- a/src/components/UncontrolledTable.js
+++ b/src/components/UncontrolledTable.js
@@ -196,6 +196,10 @@ export default class UncontrolledTable extends React.Component {
     }
   }
 
+  componentDidMount() {
+    this.props.onVisibleRowsChange(this.getVisibleRows());
+  }
+
   render() {
     const { page } = this.state;
     const { ascending, column } = this.state.sort;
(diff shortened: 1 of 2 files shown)


src/components/UncontrolledTable.js
```diff
@@ -196,6 +196,10 @@ export default class UncontrolledTable extends React.Component {
     }
   }
 
+  componentDidMount() {
+    this.props.onVisibleRowsChange(this.getVisibleRows());
```
steventhan, line 200: Can you add a test for this behavior as well? Just in case future changes won't break this

aaronmars: Yes! Actually, the tests I wrote for #943 were broken. Fixing that all up now.

steventhan (approved):
LGTM 🚢

PR #948, opened 2022-01-24
Fix CurrencyInput value attribute for browser printing

steventhan:


steventhan:
[Bug: CurrencyInput does not show value attribute when printing from browser](https://trello.com/c/dzN7cC53/58-bug-currencyinput-does-not-show-value-attribute-when-printing-from-browser)

aaronmars (approved):
LGTM.

aaronmars (approved):

PR #952, opened 2022-01-25
fix conditional rendering of child elements

kevinkir:
This fixes the case where null or false is passed as a child element to
CheckboxListInput or RadioInput. For instance:

    <CheckboxListInput>
      <FormChoice>Foo</FormChoice>
      {showBar && <FormChoice>Bar</FormChoice>}
    </CheckboxListInput>

Prior to this change, React.cloneElement would be called with null,
which would raise an exception.


kevinkir:
[Update dispatch portal checkbox styling](https://trello.com/c/RyuAgtqh/1684-update-dispatch-portal-checkbox-styling)

Peter-Lavigne (approved):

PR #954, opened 2022-01-28
fix: currencyinput should only toString if it exists

davidacevedo:
toString fails if provided with null (Seems like a test only use case). Typescript should catch this, but not everyone uses typescript yet so just providing a fix for null values.

ListItem should have selectable and onSelect as optional. This allows devs to not have to supply empty props.

For FormRow, I attempted to remove the index signature and I then get the correct type on hover
image
But then the actual parameter used thinks it is of `any` type. I want to research more on why that happens, but don't want to prevent property from updating.

We might want to update the reactstrap types for Input

### DIFF
--- a/src/components/CurrencyInput.tsx
+++ b/src/components/CurrencyInput.tsx
@@ -82,7 +82,7 @@ const CurrencyInput: FunctionComponent<Props> = ({
    * https://github.com/uNmAnNeR/imaskjs/blob/master/packages/react-imask/src/mixin.ts#L139-L147
    */
   if (value !== undefined) {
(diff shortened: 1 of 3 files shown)


ztmeyer (approved):

davidacevedo:
Code I tried to remove index from FormRow.d.ts
```
// Remove [key: any]: string from type definitions as this messes with keyof ComponentProps
type RemoveIndex<T> = {
  [ K in keyof T as string extends K ? never : number extends K ? never : K ] : T[K]
};
```

ztmeyer (approved):
🚀

PR #962, opened 2022-02-14
fix: replace custom padding with standard bootstrap

fabio-aiello-appfolio:


aaronmars:
This change has already been made in https://github.com/appfolio/react-gears/commit/55d359848e02ce11a4c9277d75846a34eabb4115

PR #963, opened 2022-02-16
fix: let BlockPanel card-header class handle padding

JeremyRH:
https://trello.com/c/FxzxQ3qZ/60-block-panel-card-header


steventhan (approved):
🚢

PR #965, opened 2022-02-17
fix: remove small uppercase style from SummaryBox

JeremyRH:
https://trello.com/c/PYD67I86/61-summarybox-label-text


davidacevedo (approved):
Looks good 👍

PR #966, opened 2022-02-17
feat(List): Allow colored backgrounds

harrypresman:
Provide the ability for each item in a list to provide the color for their
background when expanded.  The selected color (primary) will override
the expanded color.

### DIFF
--- a/src/components/List/ListItem.tsx
+++ b/src/components/List/ListItem.tsx
@@ -10,6 +10,7 @@ export interface ListItemProps<T> extends Omit<ListGroupItemProps, 'onSelect'> {
   className?: string,
   color?: string,
   expanded?: boolean,
+  expandedColor?: string,
   item: T,
   onExpand?: (item: T) => React.ReactNode | undefined,
   onSelect?: (item: T, checked?: boolean) => void
@@ -24,6 +25,7 @@ function ListItem<T>({
   className,
   color,
   expanded: defaultExpanded = false,
+  expandedColor,
   item,
   onExpand,
   onSelect,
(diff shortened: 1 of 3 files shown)


harrypresman:
[change background color of selected inquiry on new guest card](https://trello.com/c/h8Lbg2bN/492-change-background-color-of-selected-inquiry-on-new-guest-card)

davidacevedo:
I'm gonna adjust the commitlint. We just have the default one so it expects everything to be lowercase!

harrypresman:
Can we have a run script to check the commit format?  I inevitably get __something__ wrong :)

davidacevedo:
> Can we have a run script to check the commit format? I inevitably get **something** wrong :)

We will have this in the pipeline!

src/components/List/ListItem.tsx
```diff
@@ -37,9 +39,11 @@ function ListItem<T>({
   const [expanded, setExpanded] = useState(defaultExpanded);
   const [id] = useState(() => uniqueId('listitem-'));
 
+  const unselectedColor = expanded ? expandedColor : color;
```
JeremyRH, line 42: Do you think we should use `color` as a fallback in case `expandedColor` is undefined? It is a new prop and will change the behavior if not passed in.
```ts
const unselectedColor = expanded && expandedColor || color;
```

harrypresman: Well, luckily in this case, color is always undefined!  But we can certainly be more rigorous about this prop.

harrypresman:
I don't have an approval yet, any other requested changes?

davidacevedo (approved):

PR #969, opened 2022-02-22
Update eslint and format everything

aaronmars:
- Update eslint and all relevant plugins
- Add prettier and .prettierrc
- Format all files automatically using the two tools
- Make a few fixes for eslint violations that can't be fixed automatically
- Add top-level `.git-blame-ignore-revs` so the commit for the formatting can be hidden from git-blame


.prettierrc
```diff
@@ -1,4 +1,11 @@
...
-  "singleQuote": true
+  "arrowParens": "always",
+  "bracketSpacing": true,
+  "jsxSingleQuote": false,
+  "printWidth": 100,
+  "semi": true,
+  "singleQuote": true,
+  "tabWidth": 2,
+  "useTabs": false,
+  "trailingComma": "none"
```
steventhan, line 10: I've found `trailingComma` to be beneficial for reducing git diff (this pr included), do you have a strong opinion on this?

aaronmars: I don't have a *strong* opinion. I guess it's just habit, and something I'm more used to.

aaronmars: Changed to `"trailingComma": "es5"` (which is the default for prettier).
Re-formatted all files with this option in place.

steventhan (approved):
Just a minor comment, the rest looks good!

src/components/Calendar.js
```diff
@@ -134,13 +137,9 @@ class Calendar extends React.Component {
     delete props.dateEnabled; // Table doesn't need dateVisible
     delete props.dateVisible; // Table doesn't need dateVisible
 
+    /* eslint-disable react/no-array-index-key -- This should be ok for the calendar data, but */
```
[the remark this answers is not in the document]
aaronmars: Good catch. Fixed (finished)!

aaronmars:
Too much has changed since this was opened. It will be safer and easier to close and try again starting with an updated master.

PR #977, opened 2022-03-10
Convert ConfirmationButton to function/TS component

aaronmars:
Beyond the normal conversion to a function component in TS, I removed the `react-onclickoutside` dependency.
This component was the only place it was used, and I could not find an instance where this functionality was useful beyond what is provided by the `onBlur` handler. Even if the functionality was required to be put back in, a new solution would need to be developed as the `react-onclickoutside` HOC is not compatible with function components: https://github.com/Pomax/react-onclickoutside#functional-component-with-usestate-hook

The unit tests for this component were completely re-written to use `@testing-library/react` completely.


src/components/ConfirmationButton.tsx
```diff
@@ -0,0 +1,44 @@
+import noop from 'lodash.noop';
+import type { FC, MouseEventHandler, ReactNode } from 'react';
+import React, { useCallback, useState } from 'react';
+import type { ButtonProps } from './Button';
+import Button from './Button';
+
+export interface ConfirmationButtonProps extends ButtonProps {
+  children?: ReactNode;
```
steventhan, line 8: I don't think `children` is needed here

aaronmars: Oh man. That one got me again. Thanks for catching!

steventhan (approved):

PR #982, opened 2022-03-11
Bootstrap5

davidacevedo:
Update react-gears to be bootstrap5 compatible with reactstrap. The changes also reflect the bootstrap 5 migration guide https://getbootstrap.com/docs/5.0/migration/


davidacevedo:
[Update react-gears to Bootstrap 5](https://trello.com/c/wVWNQO3D/18-update-react-gears-to-bootstrap-5)

src/components/AddressInput.tsx
```diff
@@ -177,11 +177,11 @@ const AddressInput = ({
           disabled={disabled}
         />
       </FormLabelGroup>
-      <Row className="no-gutters">
+      <Row className="gx-3">
```
aaronmars, line 180: `no-gutters` was renamed to `g-0` is 3 correct here?

src/components/DateInput.js
```diff
@@ -270,7 +269,7 @@ export default class DateInput extends React.Component {
       dateFormat, defaultValue, keyboard, locale, onBlur, onChange, parse, positionFixed, value, state, ...props } = this.props; // eslint-disable-line no-shadow
     const { open } = this.state;
     const date = this.getCurrentDate();
-    const dropdownProps = open ? { positionFixed } : {};
+    const dropdownProps = open ? { strategy: positionFixed ? 'fixed' : undefined } : {};
```
aaronmars, line 272: How about this instead?
```suggestion
    const dropdownProps = open && positionFixed ? { strategy: 'fixed' } : {};
```

aaronmars (changes requested):

src/components/DateInput.js
davidacevedo: Yup, that looks better as well!

src/components/AddressInput.tsx
davidacevedo: Seems like the gutters were not necessary. Will revert to g-0

aaronmars (approved):

src/components/AddressInput.tsx
davidacevedo: Updated

src/components/DateInput.js
davidacevedo: Updated

emcpete (approved):

PR #987, opened 2022-03-15
Update eslint and format all files

aaronmars:
Updates eslint to v8.
Updates related libraries to their latest
Adds prettier and a more comprehensive `.prettierrc`
Formats all files with `prettier` and `eslint --fix`
Fixes or adds exceptions for all eslint warnings and errors that aren't auto-fixable


.eslintrc
```diff
@@ -0,0 +1,44 @@
...
+    "import/newline-after-import": "warn",
+    "import/order": ["warn", { "alphabetize": { "order": "asc" } }],
+    "max-len": "off",
+    "no-confusing-arrow": "off",
+    "no-only-tests/no-only-tests": "error",
+    "no-nested-ternary": "off",
+    "no-unused-vars": "off",
+    "react/default-props-match-prop-types": ["error", { "allowRequiredDefaults": true }],
+    "react/no-array-index-key": "warn", // TODO: Make this an error when warnings are addressed
+    "react/no-find-dom-node": "warn", // TODO: Make this an error when warnings are addressed
```
steventhan, line 23: Are these TODOs still applicable?

src/components/SortableTable.js
```diff
@@ -57,6 +71,38 @@ function defaultRenderRow(row, columns, rowClassName, rowExpanded, rowOnClick, t
...
+        id={selectRowId}
+        type="checkbox"
+        className="mx-1"
+        checked={rowSelected(row)}
+        onClick={(e) => e.stopPropagation()}
+        onChange={(e) => onSelect(row, e.target.checked)}
+      />
+    </>
+  );
+}
```
steventhan, line 91: Probably should be react component

src/components/SortableTable.js
```diff
@@ -57,6 +71,38 @@ function defaultRenderRow(row, columns, rowClassName, rowExpanded, rowOnClick, t
...
+    <Button
+      className="px-2 py-0"
+      color="link"
+      onClick={() => onExpand(row)}
+      aria-label="Expand row"
+    >
+      <Icon name={expanded ? 'angle-up' : 'angle-down'} />
+    </Button>
+  );
+}
```
steventhan, line 104: Also react component

.eslintrc
```diff
@@ -0,0 +1,44 @@
+{
```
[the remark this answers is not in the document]
aaronmars: I can make it `.json` if necessary, but `.eslintrc` is a fairly well-known filename, and the vscode (at least) has no problem knowing the format.

.eslintrc
aaronmars: I suppose not! They have been addressed in code now (either through code changes, or through `eslint-disable` comments with an explanation.

src/components/SortableTable.js
aaronmars: This was moved out of the component implementation to satisfy an eslint error. I'd definitely like to keep this PR focused on eslint stuff.
However, I definitely agree with this, so I'll do a follow-up to clean it up.

src/components/SortableTable.js
aaronmars: Same as above.

steventhan (approved):

PR #989, opened 2022-03-15
fix: better stories for popover to eliminate styling confusion

aaronmars:
The existing popover stories introduced some styling confusion as the "customized" example looked too much like the example that uses the standard `<PopoverHeader />`/`<PopoverBody />` combination.

This makes an explicit, unambiguous customized example, and leave the standard ones fairly plain.


aaronmars:
https://trello.com/c/5cDnH4iS

steventhan (approved):

PR #993, opened 2022-03-30
Migrate Calendar/MonthCalendar to TS/function

aaronmars:
- Convert Calendar and MonthCalendar to TS / function components
- Convert tests and stories to TS
- Convert tests to use `@testing-library/react`
- Break out a couple of components to their own file
- Break out a few utility functions to their own file
- Add the concept of a `*.types.ts` for use when a type/interface gets too long or is shared too many places to be inside of an implementation file.


src/components/Calendar/MonthCalendar.tsx
```diff
@@ -0,0 +1,78 @@
...
+        </Col>
+      </Row>
+    </div>
+    <style jsx>
+      {`
+        .rg-MonthCalendar {
+          min-width: 16em;
+        }
+      `}
+    </style>
```
JeremyRH, line 74: I had no idea we use `styled-jsx`!

JeremyRH (approved):

PR #995, opened 2022-03-31
Fix list issues

davidacevedo:
List was using `map.has(item)` vs the custom `hasItem` that would reference the custom `selectedKeyMapper`. This cause the selectAll button to not be able to unselect all items.

### DIFF
--- a/src/components/List/List.tsx
+++ b/src/components/List/List.tsx
@@ -119,11 +119,6 @@ function List<T extends Item>({
     }
   }, [items, Array.from(selection.values())]);
 
-  const allSelectableSelected = useMemo(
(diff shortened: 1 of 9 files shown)


src/components/List/List.tsx
```diff
@@ -119,11 +119,6 @@ function List<T extends Item>({
     }
   }, [items, Array.from(selection.values())]);
 
-  const allSelectableSelected = useMemo(
```
aaronmars, line 122: Looks like this was the only usage of `useMemo()`. Remove its import above to make the linter warnings will go away.

steventhan (approved):

aaronmars (approved):

PR #1001, opened 2022-04-04
fix: conform popover to coastline look

aaronmars:
- Change the popover story to remove the h4 from the popover body to remove confusion/conflict with the popover header.
- Force the popover header to default to an `<h4>` for its tag (rather than `<h3>` from reactstrap)

### DIFF
--- a/src/components/Popover/PopoverHeader.tsx
+++ b/src/components/Popover/PopoverHeader.tsx
@@ -1,3 +1,10 @@
-import { PopoverHeader } from 'reactstrap';
+import type { FC } from 'react';
+import React from 'react';
+import type { PopoverHeaderProps } from 'reactstrap';
+import { PopoverHeader as InternalPopoverHeader } from 'reactstrap';
+
+const PopoverHeader: FC<PopoverHeaderProps> = ({ tag = 'h4', ...props }) => (
+  <InternalPopoverHeader tag={tag} {...props} />
+);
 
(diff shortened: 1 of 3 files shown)


src/components/Popover/PopoverHeader.tsx
```diff
@@ -1,3 +1,10 @@
-import { PopoverHeader } from 'reactstrap';
+import type { FC } from 'react';
+import React from 'react';
+import type { PopoverHeaderProps } from 'reactstrap';
+import { PopoverHeader as InternalPopoverHeader } from 'reactstrap';
```
JeremyRH, line 4: [TypeScript 4.5 type modifiers on import names.](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names)
```suggestion
import { PopoverHeader as InternalPopoverHeader, type PopoverHeaderProps } from 'reactstrap';
```

aaronmars: Boom!
I'll need to fix the `@typescript-eslint/consistent-type-imports` rule to see if this syntax is supported.

aaronmars: Actually, this is a great suggestion, but I'm going to need to upgrade to TS 4.5 and then alter this syntax everywhere.
Saving all that for a follow-up PR.

PR #1004, opened 2022-04-05
Upgrade date-fns to v2

aaronmars:
- Now using `date-fns@2.28.0` ([UPGRADE GUIDE](https://date-fns.org/v2.28.0/docs/Upgrade-Guide))
- Format string changes:
  - `DD` => `d`
  - `YYYY` => `yyyy`
  - `A` => `a` (for AM/PM)
- Remove `fecha` and use `date-fns` for all parsing and formatting
- `parse()` requires a 3rd argument of type `Date` to be used as a "reference date" which defines values missing from the parsed dateString
- `parse()` now always returns a `Date`, but it might be `Invalid Date`, so I added a few calls to `isValid()` where necessary.


JeremyRH (approved):

steventhan (commented):
Even though the props are still of type `string`, this feels a lot like a breaking change, should we make it a 7 -> 8 bump to reflect this?

aaronmars:
This effort is a bit more complicated. Will follow up with specific recommendations regarding date/time parsing, formatting and utilities.

PR #1007, opened 2022-04-18
fix: allow state/country input to be empty string

clayharmon:
Setting the value to `null` causes the component to go from a controlled component to an uncontrolled component.

### DIFF
--- a/src/components/Address/CountryInput.tsx
+++ b/src/components/Address/CountryInput.tsx
@@ -6,7 +6,7 @@ import Input from '../Input/Input';
 import COUNTRIES from './util/Countries'; // TODO i18n country names based on locale
 
 interface CountryInputProps extends Omit<InputProps, 'type' | 'onChange'> {
-  onChange?: (value: string | null) => void;
+  onChange?: (value: string) => void;
   placeholder?: string;
   value?: string;
 }
@@ -31,7 +31,7 @@ const CountryInput: FC<CountryInputProps> = ({
       disabled={disabled}
       id={id}
       name={name}
-      onChange={(e) => onChange(e.target.value === '' ? null : e.target.value)}
+      onChange={(e) => onChange(e.target.value)}
     >
       <option value="">{placeholder}</option>
       {COUNTRIES.map((country) => (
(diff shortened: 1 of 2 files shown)


src/components/Address/CountryInput.tsx
```diff
@@ -31,7 +31,7 @@ const CountryInput: FC<CountryInputProps> = ({
       disabled={disabled}
       id={id}
       name={name}
-      onChange={(e) => onChange(e.target.value === '' ? null : e.target.value)}
+      onChange={(e) => onChange(e.target.value)}
```
JeremyRH, line 34: The type `CountryInputProps` above needs to be changed:
```diff
interface CountryInputProps extends Omit<InputProps, 'type' | 'onChange'> {
-  onChange?: (value: string | null) => void;
+  onChange?: (value: string) => void;
  placeholder?: string;
  value?: string;
}
```

src/components/Address/StateInput.tsx
```diff
@@ -56,7 +56,7 @@ const StateInput: FC<StateInputProps> = ({
       disabled={disabled}
       id={id}
       name={name}
-      onChange={(e) => onChange(e.target.value === '' ? null : e.target.value)}
+      onChange={(e) => onChange(e.target.value)}
```
JeremyRH, line 59: Same change needs to be made to `StateInputProps`.

JeremyRH (changes requested):
I think [AddressInput](https://github.com/appfolio/react-gears/blob/bbac01a3a692486fb3542c7e79eed60c531c5968/src/components/Address/AddressInput.tsx#L48) relies on `null` for showing default values.
We should make sure it works as expected.

clayharmon:
@JeremyRH thanks for your review, after looking more into this it seems that in the latest version this is not an issue, we are on `6.11.0` which is showing this error.

JeremyRH:
@clayharmon Darn. I'm always glad to see someone remove unintuitive behavior.
I understand though. This would be a small breaking change and needs a bit of work to test.

PR #1009, opened 2022-04-25
Allow RadioGroup label to be JSX element

JeremyRH:
The label was used to generate the ID. A JSX element isn't going to work for this so I removed the ID.
Should be fine because the ID had a randomly generated string in it and most likely was not used outside this component.


aaronmars (approved):

xl5liu (approved):
LGTM! Thanks for making the change!

PR #1026, opened 2022-05-12
Merge react-gears-tree to react-gears

limbov8:
Merge the `Tree` component from `react-gears-tree` to `react-gears`.

### DIFF
--- a/src/components/Tree/Tree.spec.js
+++ b/src/components/Tree/Tree.spec.js
@@ -1,0 +1,88 @@
+import React from 'react';
+import { render, screen, fireEvent } from '@testing-library/react';
+import Tree from './Tree';
+
+describe('<Tree />', () => {
+  let data;
+  beforeEach(() => {
+    data = [
+      {
+        item: 'Pepperoni',
+        expanded: true,
+        selected: null,
+        children: [
+          {
+            item: 'Spicy',
+            expanded: false,
+            selected: false,
+          },
+          {
+            item: 'Regular',
+            expanded: false,
+            selected: true,
+          },
+        ],
+      },
+      {
+        item: 'Chicken',
+        expanded: false,
+        selected: false,
+        children: [
+          {
+            item: 'Buffalo',
+            expanded: false,
+            selected: false,
+            children: [
+              {
+                item: 'Mild',
+                expanded: false,
+                selected: false,
+              },
+              {
+                item: 'Hot',
+                expanded: false,
+                selected: false,
+                children: [
+                  {
+                    item: 'Jalapeño',
+                    expanded: false,
+                    selected: false,
+                  },
+                  {
+                    item: 'Cayenne',
+                    expanded: false,
+                    selected: false,
+                  },
+                ],
+              },
+            ],
+          },
+          {
+            item: 'BBQ',
+            expanded: false,
+            selected: false,
+          },
+        ],
+      },
+    ];
+  });
+
+  it('should render correctly', () => {
+    const labelRenderer = (item, { selected }) => (
+      <div className="d-flex align-items-center" data-testid="js-item-label">
+        <h4 className="m-0 font-weight-normal">{item}</h4>
+        <span className="mx-2 text-secondary">
+          {selected ? `${item} is selected` : `${item} is not selected`}
+        </span>
+      </div>
+    );
+    const updateOption = jest.fn();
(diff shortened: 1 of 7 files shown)


limbov8:
[Bump the @appfolio react packages in partner portal](https://trello.com/c/HwRULzsa/1711-bump-the-appfolio-react-packages-in-partner-portal)

.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 12.18.3
```
davidacevedo, line 1: Can this be the latest node 14 instead.

davidacevedo: Nevermind, you can delete this as this PR already solves that https://github.com/appfolio/react-gears/pull/1027/files

src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,92 @@
+import classnames from 'classnames';
+import React, { useEffect, useRef, useState } from 'react';
+import { v4 as uuidv4 } from 'uuid';
```
davidacevedo, line 3: Instead of uuid, can you follow the getId pattern found here https://github.com/appfolio/react-gears/pull/1027/files

aaronmars (commented):
Could we rename Options.ts to Tree.types.ts? This is a pattern one or two other places in this library.

Would it be possible for any unit testing?

davidacevedo:
> Could we rename Options.ts to Tree.types.ts? This is a pattern one or two other places in this library.
> 
> Would it be possible for any unit testing?

@aaronmars what about adding the type to `Tree.tsx`. It's only being used by that file.

aaronmars:
> what about adding the type to `Tree.tsx`. It's only being used by that file.

For sure! Keeping it local is fine too. That seems like the logical thing to do in this case.
The separate types file is fine where it is necessary or cleaner.

limbov8:
Actually, the `Options.ts` has been used in `Tree.tsx` and `TreeItem.tsx`. I have changed the file name to `Tree.types.ts`

aaronmars:
@limbov8: Regarding the types file, that sounds great. Sorry I missed that extra usage earlier.
I noticed that the old, standalone repo had nothing in the way of tests, but do you think it would be possible to add a `.spec.tsx` file or 2 so something can be done?
Just a basic `render()` test using `@testing-library/react` would be great for starters.

src/components/Tree/TreeItem.spec.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import sinon from 'sinon';
+import assert from 'assert';
+import { shallow } from 'enzyme';
```
davidacevedo, line 4: Hey Libo, for tests we are trying to use `@testing-library/react` now. https://testing-library.com/docs/react-testing-library/intro/ You can find a few examples of how to use it in the react-gears code base

src/components/Tree/TreeItem.spec.js
```diff
@@ -0,0 +1,42 @@
...
+      updateOption: sinon.stub()
+    };
+  });
+
+  afterEach(() => sinon.restore());
+
+  it('should render correctly', () => {
+    const wrapper = shallow(<TreeItem {...props} />);
+
+    assert(wrapper);
```
davidacevedo, line 35: Also, instead of assert we are using the jest `expect` now. It should be available with no imports.

src/components/Tree/TreeItem.spec.js
limbov8: Will update. What's the rational on migrating to testing-library/react from enzyme?

src/components/Tree/Tree.spec.js
```diff
@@ -0,0 +1,90 @@
...
+      </div>
+    );
+    const updateOption = jest.fn();
+    render(<Tree options={data} label={labelRenderer} onChange={updateOption} selectable />);
+
+    expect(screen.queryAllByTestId('js-item-label').length).toEqual(10);
+    expect(screen.queryByText('Regular is selected')).toBeInTheDocument();
+    expect(screen.queryByText('Chicken is not selected')).toBeInTheDocument();
+    fireEvent.click(
+      screen.getByText('Spicy').parentElement.parentElement.parentElement.querySelector('input')
```
davidacevedo, line 86: `data-testid` with a `queryAll` selector might've been better there, but we can take on that kind of update as there was no other good way to find this element.

davidacevedo (approved):
Looks good 👍 Thanks for the patience!

PR #1027, opened 2022-05-13
Replace uniqid dependency with own module

tkissingappfolio:
uniqid tries to require node-specific modules to allow better random
strings in environments that allow access to hardware information.
These includes require additional mocking/configuration for consumers of
react-gears without any benefit, because react-gears runs on the client.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,0 +1,1 @@
+nodejs lts-erbium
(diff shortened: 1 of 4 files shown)


davidacevedo:
Mind converting to something like this https://github.com/appfolio/react-gears/blob/74882a4339c68580694835481a837f400c3877e2/src/components/TruncatedText/TruncatedText.tsx#L16. This is found more throughout the code, so I'd rather it just follow current patterns for now.

tkissingappfolio:
Changed to a simple counter as requested

davidacevedo (approved):

PR #1029, opened 2022-05-16
Upgrade date-fns to v2

daviditsygin:
Updated the `date-fns` dependency and its usages.


daviditsygin:
[Update react-gears to date-fns 2](https://trello.com/c/F145JQh6/394-update-react-gears-to-date-fns-2)

JeremyRH:
Another attempt to update date-fns here: https://github.com/appfolio/react-gears/pull/762

Unfortunately, the format string is a prop for some components meaning we don't have control of all the inputs to data-fns. This is a major breaking change for these components. It's easier to just re-write them in a new library and convert usage of these components to the new ones.

PR #1031, opened 2022-05-23
New combo select box

aaronmars:
Create a new component called `<MultiSelectCombobox>` that is intended to replace usages of the existing `<Combobox multi>` 
This is a new, separate component from Combobox, and has some building blocks that could be used for a rewrite of it in the future, if that is found to be beneficial for simplicity or architectural reasons.

There is one top-level component here. The `<MultiSelectCombobox>` can work in either controlled or uncontrolled mode. If children are supplied, the component is completely controlled and the lists of selections and items are handled externally:
```jsx
<MultiSelectCombobox>
  <ComboboxSelections>One</ComboboxSelection>
    <ComboboxSelection>Three</ComboboxSelection>
    <ComboboxSelection>Five</ComboboxSelection>
  </ComboboxSelections>
  <FilteredComboboxItems>
    <ComboboxItem>Two</ComboboxItem>
    <ComboboxItem>Four</ComboboxItem>
  </FilteredComboboxItems>
</MultiSelectCombobox>
```
If no children are supplied, the component expects the `options` and `onChange` props to be supplied. In this case, the selections and visible options are handled internally by the component's local state:
```jsx
const options = [
  { label: 'One', value: 'one' },
  { label: 'Two', value: 'two' },
  { label: 'Three', value: 'three' },
];

<MultiSelectCombobox options={options} onChanged={(sel) => { console.log('Selected items:', sel) } />
```


src/components/MultiSelectCombobox/ComboboxItem.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+interface ComboboxItemProps {
+  onClick: MouseEventHandler;
+}
+
+const ComboboxItem: FC<ComboboxItemProps> = ({ children, onClick }) => (
+  <DropdownItem
+    onClick={onClick}
+    className="overflow-hidden text-truncate"
+    title={children as string}
```
steventhan, line 13: Would it be a problem if `children` is passed in as a non-string react component? If it is, I think we should narrow down the `children` type for `ComboboxItemProps`

```diff
interface ComboboxItemProps {
  onClick: MouseEventHandler;
+ children: string;
}
```

src/components/MultiSelectCombobox/ComboboxSelection.tsx
```diff
@@ -0,0 +1,42 @@
...
+    onRemove();
+  };
+  return (
+    <Badge
+      className="selection d-flex flex-nowrap align-items-center m-1 user-select-none"
+      color="secondary"
+      style={{ minWidth: '0' }}
+      onClick={handleRemoveClick}
+      role="listitem"
+      title={children as string}
```
steventhan, line 28: Shall we narrow `children` type here as well?

src/components/MultiSelectCombobox/ComboboxSelections.tsx
```diff
@@ -0,0 +1,61 @@
...
+interface ComboboxSelectionsProps {
+  onRemoveAll?: () => void;
+  placeholder?: string;
+}
+
+const defaultProps = { placeholder: 'Click to select an option...' };
+
+const ComboboxSelections: FC<ComboboxSelectionsProps> = ({
+  children,
+  onRemoveAll,
```
steventhan, line 16: Nit: make this `lodash.noop` as well for consistency

src/components/MultiSelectCombobox/UncontrolledMultiSelectCombobox.tsx
```diff
@@ -0,0 +1,172 @@
...
+import React, { useCallback, useEffect, useRef, useState } from 'react';
+import type { ComboboxOption } from './Combobox.types';
+import ComboboxItem from './ComboboxItem';
+import ComboboxItems from './ComboboxItems';
+import ComboboxSelection from './ComboboxSelection';
+import ComboboxSelections from './ComboboxSelections';
+import FilteredComboboxItems from './FilteredComboboxItems';
+import MultiSelectCombobox from './MultiSelectCombobox';
+
+type OptionComparisonKey = 'id' | 'value';
```
steventhan, line 11: Duplicate type I assume, there's an identical one in `Combobox.types.ts`

src/components/MultiSelectCombobox/UncontrolledMultiSelectCombobox.tsx
```diff
@@ -0,0 +1,172 @@
...
+function isOptionFiltered<T extends ComboboxOption>(option: T, filterText: string) {
+  return filterText === '' || option.label.toLowerCase().includes(filterText.toLowerCase());
+}
+
+const defaultProps = {
+  allowCreation: false,
+  closeOnSelect: false,
+  filterOptions: true,
+  onChange: noop,
+  optionComparisonKey: 'value' as OptionComparisonKey,
```
steventhan, line 40: I think this also works if you want to type fewer chars 😄  
```suggestion
  optionComparisonKey: 'value' as const,
```

src/components/MultiSelectCombobox/ComboboxItem.tsx
aaronmars: Good point. I'm actually going to narrow down a few of the other `children` types as well.

src/components/MultiSelectCombobox/ComboboxSelections.tsx
aaronmars: Instead of adding `noop` to props, I think it's a better pattern to just call these event functions like:
```javascript
onRemoveAll?.();
```
(which is just a nice way of doing the old, busted `onRemoveAll && onRemoveAll()`)

Any issues with that? Seems to be the cleanest.

steventhan: Yep, no strong opinion on this, either way is fine, just thought since other components in this PR use `noop`, might as well make this the same or vice versa. Another option (also my personal favorite) is `onRemoveAll = () => undefined`, basically `noop` without the import since `undefined` fits under `void`

src/components/MultiSelectCombobox/ComboboxItem.tsx
```diff
@@ -0,0 +1,20 @@
...
+export interface ComboboxItemProps {
+  children: string;
+  onClick: MouseEventHandler;
+}
+
+const ComboboxItem = ({ children, onClick }: ComboboxItemProps) => (
+  <DropdownItem
+    onClick={onClick}
+    className="overflow-hidden text-truncate"
+    title={children as string}
```
steventhan, line 14: Can we remove the type casting here?

src/components/MultiSelectCombobox/ComboboxSelection.tsx
```diff
@@ -0,0 +1,37 @@
...
+    onRemove?.();
+  };
+  return (
+    <Badge
+      className="selection d-flex flex-nowrap align-items-center m-1 user-select-none"
+      color="secondary"
+      style={{ minWidth: '0' }}
+      onClick={handleRemoveClick}
+      role="listitem"
+      title={children as string}
```
steventhan, line 23: The casting is probably not needed here either

steventhan (approved):
Minor comments, but LGTM!

src/components/MultiSelectCombobox/ComboboxItem.tsx
aaronmars: Yes! I just saw that one myself. Now that `children` is `string`, that can go. Much better.

src/tooling/comboboxData.ts
```diff
@@ -0,0 +1,63 @@
...
+  { label: 'U.S. Armed Forces Europe', value: 'AE' },
+  { label: 'U.S. Armed Forces Pacific', value: 'AP' },
+  { label: 'American Samoa', value: 'AS' },
+  { label: 'Micronesia', value: 'FM' },
+  { label: 'Guam', value: 'GU' },
+  { label: 'Marshall Islands', value: 'MH' },
+  { label: 'Northern Mariana Islands', value: 'MP' },
+  { label: 'Puerto Rico', value: 'PR' },
+  { label: 'Virgin Islands', value: 'VI' },
+];
```
JeremyRH, line 63: Is this used?

JeremyRH (approved):

src/tooling/comboboxData.ts
aaronmars: Good catch. That got dropped when I refactored for the new hybrid abstraction.
I added its usage back in as I think it's good to have an example with a long list of options.

PR #1043, opened 2022-07-18
Add option to center years on MonthInput and MonthCalendar

kkirasich:
If you do not pass a `value` to MonthInput, the year selector on MonthCalendar is this year through the previous 12 years. 

image

Our PM got tripped up in a demo trying to schedule something for January the following year but found the component difficult to navigate to the next year.  This PR allows users to center the years such that whatever `value` is passed in or defaulted, that year will be in the center on the available year inputs.

image

### DIFF
--- a/src/components/Input/MonthInput.spec.js
+++ b/src/components/Input/MonthInput.spec.js
@@ -250,8 +250,10 @@ describe('<MonthInput />', () => {
     const callback = sinon.spy();
     const defaultDate = new Date(2017, 7, 14);
     const dateVisible = (date) => isSameDay(date, defaultDate);
+    const centerYearSelection = true;
     const component = mount(
       <MonthInput
+        centerYearSelection={centerYearSelection}
         defaultValue={defaultDate}
         onChange={callback}
         dateVisible={dateVisible}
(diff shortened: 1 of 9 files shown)


gcpreston (approved):
Big fan of adding to `stories` files too

src/components/Input/MonthInput.spec.js
```diff
@@ -266,6 +268,11 @@ describe('<MonthInput />', () => {
       assert.equal(calendar.props().dateVisible, dateVisible);
     });
 
+    it('should pass centerYearSelection to Calendar component', () => {
+      const calendar = component.find('MonthCalendar');
+      assert.equal(calendar.props().centerYearSelection, centerYearSelection);
+    });
```
JeremyRH, line 274: Is there any use for this test? It seems like it's only asserting the setup of the test.

aaronmars (approved):

src/components/Input/MonthInput.spec.js
kkirasich: Not needed, just saw another test that tests prop pass-through

JeremyRH (approved):

PR #1048, opened 2022-07-26
feat: make MultiSelectCombobox composable, allow input of selections

aaronmars:
- Add types entries for the composition components
- Rename `ControlledMultiSelectCombobox` to `ComboboxWrapper` to better describe its role in the composition.
- Show an example of custom Item rendering using composition in the Uncontrolled case (satisfies https://trello.com/c/IyMBLsiB)
- Add composition semantics in `UncontrolledMultiSelectCombobox` 
- Add prop to pre-select selections in `UncontrolledMultiSelectCombobox` 
- Add a few keyboard tests that were missed in the initial implementation

### DIFF
--- a/src/components/MultiSelectCombobox/UncontrolledMultiSelectCombobox.tsx
+++ b/src/components/MultiSelectCombobox/UncontrolledMultiSelectCombobox.tsx
@@ -1,16 +1,23 @@
 import React, { useCallback, useEffect, useRef, useState } from 'react';
-import type { ComboboxOption, OnChangeEnventHandler } from './Combobox.types';
-import ComboboxItem from './ComboboxItem';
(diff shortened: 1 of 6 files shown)


src/components/MultiSelectCombobox/UncontrolledMultiSelectCombobox.tsx
```diff
@@ -30,22 +37,39 @@ function isOptionFiltered<T extends ComboboxOption>(option: T, filterText: strin
 }
 
 const defaultProps = {
+  components: {},
```
JeremyRH, line 40: nit: I don't think you need an empty object as a default prop. Object spread is fine for `undefined`:
```js
{ ...undefined } // {}
```

JeremyRH (approved):

PR #1052, opened 2022-08-09
chore: convert to yarn

aaronmars:


package.json
```diff
@@ -22,28 +23,28 @@
   "scripts": {
     "tslint": "tsc --noEmit",
     "eslint": "eslint src --ext .js,.ts,.jsx,.tsx",
-    "eslint:fix": "npm run eslint -- --fix",
-    "lint": "npm run tslint && npm run eslint",
-    "lint:ts": "npm run dist",
+    "eslint:fix": "yarn eslint -- --fix",
```
steventhan, line 26: Super minor: I think we can leave the `--` out

```suggestion
    "eslint:fix": "yarn eslint --fix",
```

.github/workflows/test.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
-      - run: npm ci
-      - run: npm run lint
-      - run: npm run format
-      - run: npm test
+      - run: yarn install --immutable
```
steventhan, line 26: I think within gh action, the `CI` env variable should be set, so `yarn install` should already been run in `--immutable` mode

steventhan (approved):
Minor comments, but everything looks good to me 🚢

PR #1054, opened 2022-08-13
Convert FormRow and FormLabelGroup to TS

JeremyRH:
No breaking changes except to the types. The prop `width` was wrong.


JeremyRH:
[Convert FormRow & FormLabelGroup to TypeScript](https://trello.com/c/4JLAHgWt/135-convert-formrow-formlabelgroup-to-typescript)

emcpete (approved):
Looks good

src/components/Checkbox/CheckboxListInput.tsx
```diff
@@ -1,15 +1,22 @@
...
+  value?: string[];
+}
+
+class CheckboxListInput extends React.Component<CheckboxListInputProps> {
   static propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
     onChange: PropTypes.func,
     value: PropTypes.array,
   };
```
davidacevedo, line 17: Do we still want to keep PropTypes when converting to typescript?

JeremyRH: Probably not worth it because it only provides console errors when running dev builds. So it's usually just noise.
I didn't spend time refactoring this file because it wasn't originally planned to be converted to TS, I was forced to due to manual .d.ts types making it into generated .d.ts files. The TS compiler doesn't know we copy .d.ts file from src to the build so it errors saying it can't link to the type.

src/util/bindProps.tsx
```diff
@@ -0,0 +1,32 @@
...
+) {
+  type OProps = ComponentPropsWithoutRef<C>;
+  type ORefInstance = ElementRef<C>;
+  type BProps = OmitByType<P, undefined>;
+  const Comp = Component as ComponentType<OProps>;
+
+  return forwardRef(
+    (props: Omit<OProps, keyof BProps> & Partial<BProps>, ref: ForwardedRef<ORefInstance>) => {
+      const allProps = { ...boundProps, ...props } as OProps;
+      return <Comp {...allProps} ref={ref} />;
```
davidacevedo, line 29: Most component in reactstrap don't actually have a ref prop. It usually uses `innerRef`

JeremyRH: `ref` will be undefined if not used so it shouldn't hurt. `innerRef` will be passed in through normal props.

PR #1059, opened 2022-08-25
chore: remove unnecessary props from FeedbackButtton

ioneone:
`recipient`, `cc`, and `feature` are just passed to the callback of `onSubmit`. Since we require `onSubmit`, the client already has access to `recipient`, `cc`, `feature`.


JeremyRH:
This is a major breaking change. There are many usages of these props and removing them will require many changes in apm_bundle: https://cs.github.com/?scopeName=All+repos&scope=&q=repo%3Aappfolio%2Fapm_bundle+%3CFeedBackButton

If you are willing to fix all the cases where these props are used, we can merge this.

jmaher409 (approved):

PR #1060, opened 2022-08-25
Swap arrow direction in List->SortHeader

aaronmars:
The arrow orientation should be:
- ascending: up arrow
- descending: down arrow

This change fixes that for the List SortHeader component


aaronmars:
### Ascending => UP
ascending_up

### Descending => DOWN
descending_down

JeremyRH (approved):

PR #1088, opened 2022-09-16
add props initiallyHiddenTimes to TimeInput

taowen19:
- Updated function selectedOptions to support clear input programmatically.
- Added prop initiallyHiddenTimes to hide times when users haven't input anything.

```
    <TimeInput
      initiallyHiddenTimes={['01:30']}
    />

```

### DIFF
--- a/src/components/Input/TimeInput.js
+++ b/src/components/Input/TimeInput.js
@@ -93,6 +93,7 @@ export default class TimeInput extends React.Component {
     noResultsText: PropTypes.string,
     step: PropTypes.number, // TODO? 1-60
     timeFormat: PropTypes.string,
+    initiallyHiddenTimes: PropTypes.array,
     value: PropTypes.string,
   };
 
@@ -177,8 +178,8 @@ export default class TimeInput extends React.Component {
    * - missing whitespace "9:30AM"
    * - typing am/pm upper or lower case "9:30 am"
    */
-  filterOption = ({ label, time }, input) => {
-    const { step } = this.props;
+  filterOption = ({ label, time, value }, input) => {
+    const { step, initiallyHiddenTimes } = this.props;
 
     const removeWhitespace = (str) => str.replace(/\s/gi, '');
     const removeLeadingZeros = (str) => str.replace(/^0*/, '');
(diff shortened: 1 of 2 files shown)


taowen19:
[When selecting/changing times, user shouldn't be able to view times that are already booked](https://trello.com/c/SAVcRZeg/3504-when-selecting-changing-times-user-shouldnt-be-able-to-view-times-that-are-already-booked)

src/components/Input/TimeInput.js
```diff
@@ -187,6 +187,10 @@ export default class TimeInput extends React.Component {
 
     const [hasTypedTens, hasTypedMin] = userInputProgress(inputCandidate, time);
 
+    if (input === '' && unavailableTimes?.has(label)) {
+      return false;
+    }
```
darreneng, line 192: What's the desired behavior here? This will hide the time **only** when the user hasn't typed anything yet. Once the user starts typing, they can still enter the time?

darreneng: Generally speaking, I would suggest checking against the `value` instead of the `label`

taowen19: > What's the desired behavior here? This will hide the time **only** when the user hasn't typed anything yet. Once the user starts typing, they can still enter the time?

yes, that's the expected behavior.

darreneng: Honestly it seems a little odd to me to support this behavior here because this use case is so specific and this is meant to be a fairly reusable component.

Personally I think one way forward for this is to split up this component into multiple functions / hooks and then expose those as part of this package and reuse them elsewhere.


If we were to reuse this component maybe a prop for determining the initially displayed times would make more sense? This isn't really "unavailable" times because the user can still select them.

taowen19: Yes I can see the prop name can be a little confusing, we could change it to be 'initiallyHiddenTimes' if that sounds more clear. we could also add notes/comments.

src/components/Input/TimeInput.js
```diff
@@ -187,6 +187,10 @@ export default class TimeInput extends React.Component {
 
     const [hasTypedTens, hasTypedMin] = userInputProgress(inputCandidate, time);
 
+    if (input === '' && unavailableTimes?.has(label)) {
```
aaronmars, line 190: I'm assuming `unavailabeTimes` is a Set based on the tests and the use of `has()`.

I feel like an array would be a better option here. Also, it would help to include it in `propTypes` so there's no ambiguity.

taowen19: Updated.

aaronmars (approved):

PR #1093, opened 2022-09-27
fix: prevent list item content from overflowing

brandon-nadell44:
## Current
Column width is unbounded, so the row content overflows, causing columns to be misaligned and the horizontal scroll bar appearing
Screen Shot 2022-09-23 at 1 34 06 PM

## Fixed
Set a max-width for a div in `ListItem` containing it's content
Screen Shot 2022-09-27 at 4 00 01 PM

### DIFF
--- a/src/components/List/ListItem.tsx
+++ b/src/components/List/ListItem.tsx
@@ -54,7 +54,7 @@ function ListItem<T>({
     >
       <div className="d-flex align-items-center">
         {select && (
-          <div className="pe-2">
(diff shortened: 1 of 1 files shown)


brandon-nadell44:
[Unmatched bank transaction will indent if description is too long](https://trello.com/c/MzgKIISa/194-unmatched-bank-transaction-will-indent-if-description-is-too-long)

davidacevedo (approved):
LGTM 👍

PR #1095, opened 2022-10-03
feat: allow item to be selectable by clicking anywhere on the item

brandon-nadell44:
## Overview

Added an optional prop to `List` that allows the whole row to be clickable, instead of just on the checkbox/radio button.

Edit: simplified code so explanation below is outdated
<strike>
## Some explanation to why I implemented it like this:

- As defined by bootstrap, there must be a `Label` component wrapping everywhere that is clickable. This is why I have `Label` wrapping the entire main `div`, instead of having `Label` and `Input` as siblings (how it used to be)
  - When `Label` is wrapping an `Input`, we must explicitly change the pointer behavior
- `ListGroupItem` must still be the outermost component in order to be the child of `ListGroup`, otherwise the item borders aren't correct
  - Then, we remove the padding from `ListGroupItem` and instead put it on `Label`, so that this padding is also clickable
</strike>

### DIFF
--- a/src/components/List/List.tsx
+++ b/src/components/List/List.tsx
@@ -5,6 +5,7 @@ import { ListGroupProps } from 'reactstrap';
 import useDeepCompareEffect from 'use-deep-compare-effect';
 import useMap from '../../hooks/useMap';
 import Input from '../Input/Input';
+import Label from '../Label/Label';
 import Col from '../Layout/Col';
 import Row from '../Layout/Row';
 import ScrollContainer from '../ScrollContainer/ScrollContainer';
@@ -45,6 +46,7 @@ export interface ListProps<T> extends Omit<ListGroupProps, 'onSelect'> {
   sortByLabel?: SortHeaderProps['sortByLabel'];
   sortOptions?: SortHeaderProps['sortOptions'];
   selectable?: (item: T) => boolean;
+  selectAnywhere?: boolean;
 }
 
 const defaultProps = {
@@ -58,6 +60,7 @@ const defaultProps = {
   sort: {},
   sortByLabel: 'Sort by',
   selectable: () => true,
+  selectAnywhere: false,
 };
 
 function List<T extends Item>({
@@ -81,6 +84,7 @@ function List<T extends Item>({
   sortByLabel = defaultProps.sortByLabel,
   sortOptions,
   selectable = defaultProps.selectable,
+  selectAnywhere = defaultProps.selectAnywhere,
   ...props
 }: ListProps<T>) {
   const {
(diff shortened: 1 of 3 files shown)


brandon-nadell44:
[Allow clicking on the whole row to select](https://trello.com/c/Jyo2exC1/238-allow-clicking-on-the-whole-row-to-select)

src/components/List/List.tsx
```diff
@@ -45,6 +46,7 @@ export interface ListProps<T> extends Omit<ListGroupProps, 'onSelect'> {
   sortByLabel?: SortHeaderProps['sortByLabel'];
   sortOptions?: SortHeaderProps['sortOptions'];
   selectable?: (item: T) => boolean;
+  selectAnywhere?: boolean;
```
almgong, line 49: Would it be valuable to add an entry to `defaultProps` ?

```
const defaultProps = {
  ...,
  selectAnywhere: false
};
```

src/components/List/List.tsx
```diff
@@ -160,6 +163,10 @@ function List<T extends Item>({
 
   const showHeader = header || select === 'checkbox' || select === 'switch' || onFilter || onSort;
 
+  const selectAnywhereProps = selectAnywhere
+    ? { tag: Label, style: { cursor: 'pointer', marginBottom: 0 } }
```
almgong, line 167: Could you elaborate on why the 0 margin bottom is needed?

src/components/List/List.tsx
```diff
@@ -13,6 +13,7 @@ import SortHeader, { SortHeaderProps } from './components/SortHeader';
 import ListGroup from './ListGroup';
 import ListGroupItem from './ListGroupItem';
 import ListItem, { ListItemProps } from './ListItem';
+import Label from '../Label/Label';
```
almgong, line 16: Let's move this per the warning 👍

almgong (commented):
Had a few questions - also should we add test coverage for these changes?

src/components/List/List.tsx
brandon-nadell44: I didn't find that this was necessary (I think omitted props default to false anyways?)

src/components/List/List.tsx
brandon-nadell44: The Label tag adds bottom margin styling, so I make it 0 to be consistent with a non-selectAnywhere row

src/components/List/List.tsx
brandon-nadell44: That's weird, the warning only shows when I view all changes so I totally missed it

src/components/List/List.tsx
almgong: I think they default to `undefined`, but this would help document that the datatype of `selectAnywhere` is a boolean.

almgong (approved):

brandon-nadell44:
> Can you add a story to these changes or a story knob? I'd like to see what the changes accomplish and for what type of UI

Sure, I added a knob `selectAnywhere` to component List -> With Everything

davidacevedo (approved):
👍

aaronmars:
We're going to close this to focus on another solution in the new List over in react-components.
According to @brandon-nadell44 , it's no longer a priority for his team.

PR #1101, opened 2022-10-27
chore: bump reactstrap

fabio-aiello-appfolio:
### Bump reactstrap version to 9.1.5

Necessary to include reactstrap fix for multiple open modals that did not restore page scroll on exit.


emcpete (approved):
Looks good

PR #1107, opened 2022-11-05
feat: add HasManyFields2

jupp0r:
This change adds HasManyfields2, a successor to HasManyFields that is more composable and easier to integrate with state management libraries like mobx and easier to adopt to server-side rendering and dynamic hydration (compared to HasManyfields).


aaronmars:
Might this be more appropriate over in `react-libs` (the 'components' project within that monorepo).

aaronmars:
> Can somebody use it right now it APM? 

For sure! I have yet to do it, but I'm in the middle of the first migration of `react-gears` List to `react-components` List. 
The library should be all ready to go, though.
One benefit (if you choose to exercise it) is that you can keep the name as `HasManyFields` since it's in a completely new library.
I can definitely help with the move from `react-gears` to `react-components`

aaronmars:
Closing as "won't do". This use case will become part of the Coastline design system, and handled through that process.

PR #1109, opened 2022-11-17
fix: dateInput onClose not called with last picked date

frmalinowski:
DateInput onClose callback is not called with last picked date when value property is present


src/components/Input/DateInput.js
```diff
@@ -274,7 +274,7 @@ export default class DateInput extends React.Component {
   componentDidUpdate(prevProps, prevState) {
     this.setInputValue();
     if (this.props.onClose && this.state.open !== prevState.open && !this.state.open) {
-      const value = this.props.value !== undefined ? this.props.value : this.state.value;
+      const value = this.state.value;
```
JeremyRH, line 277: `this.state.value` can be an empty string even if `props.value` is provided. Try it by opening the calendar and closing it without selecting a date and providing `props.value`.

JeremyRH: I think we can fix the issue I mentioned above by changing this line in the contructor:
```diff
- let value = props.defaultValue || '';
+ let value = props.value || props.defaultValue || '';
```

PR #1110, opened 2022-11-20
wa-216 update ellipsis and times icons to font awesome v6.2.1

Waqas84:


fabio-aiello-appfolio:
[Revert React-Gears to fontAwesome 5](https://trello.com/c/kHEoW73y/239-revert-react-gears-to-fontawesome-5)

PR #1120, opened 2022-12-17
Enhancement: enable dynamic Combobox options (i.e. type-to-search)

xeger:
My project is using `react-gears@5` which has an early iteration of `Combobox`. I want to enhance it, so I've already forked the component from `react-gears@7` into my project (hooray, modular libraries!) so I can benefit from the newer `Combobox` UX.

My goal with this PR is to illustrate my enhancements and maybe get them accepted upstream, so I can get rid of my fork when I upgrade to `react-gears@7`.

The enhancement proposal is tiny: I've added an `onChangeText` event that fires when the user is typing into the input, irrespective of which options exist or how the typing is refining the display of options. This gives parent components the chance to go find matching options, then update the `options` prop to display matching options.

In other words: this enhancement lets me use `Combobox` to build a search-select component that finds matching options on the server as the user types keystrokes.

The event name probably isn't quite right, since the user can pass any input `type` to `Combobox` (not just `text`). Naming choices aside, I think the idea is sound because it enables a powerful higher-level behavior without baking all of that complexity into the core component.

There are some usability issues with dynamic options (i.e. flicker) that I may need to address, but I'm happy to do that later if the team thinks this is the right direction.

### DIFF
--- a/src/components/Combobox/Combobox.spec.js
+++ b/src/components/Combobox/Combobox.spec.js
@@ -80,6 +80,18 @@ describe('<Combobox />', () => {
     sinon.assert.calledWith(mockOnChange, OPTIONS[2].value);
   });
 
+  it('should call onChangeText when user types', () => {
+    const mockOnChangeText = sinon.spy();
+    const combobox = render(<Combobox options={OPTIONS} onChangeText={mockOnChangeText} />);
+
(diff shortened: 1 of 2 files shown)


xeger:
Closing, as this has been done differently in `react-components`.

PR #1129, opened 2023-01-25
Fix Select defaultValue

JeremyRH:
Don't `useEffect` to sync props with state: https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

### DIFF
--- a/src/components/Select/Select.js
+++ b/src/components/Select/Select.js
@@ -1,5 +1,4 @@
 import classnames from 'classnames';
-import noop from 'lodash.noop';
 import PropTypes from 'prop-types';
 import React, { useState } from 'react';
 import ReactSelect from 'react-select-plus';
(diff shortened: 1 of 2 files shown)


src/components/Select/Select.js
```diff
@@ -29,7 +29,9 @@ const Select = ({
 }) => {
   const [value, setValue] = useState(valueProp || props.defaultValue);
 
-  useEffect(() => setValue(valueProp), [valueProp]);
+  if (typeof valueProp !== 'undefined' && !Object.is(valueProp, value)) {
+    setValue(valueProp);
+  }
```
infiniteluke, line 34: This works, but I think we should rather use the `defaultValue` prop on `Select`. I try to avoid setting state inside the render path if at all possible. I'm pretty sure `react-select-plus` supports `defaultValue`, right?

Something like:

```jsx
let valProps = typeof valueProp !== 'undefined' ? { value: valueProp } : { defaultValue };

return (
  <SelectElement
    {...valProps}
    arrowRenderer={({ isOpen }) => getSelectArrow(isOpen, arrowRenderer)}
    clearRenderer={getCloseButton}
    optionComponent={Option}
    inputProps={{ name, ...inputProps }}
    // ... other props
 />
)
```

JeremyRH: > react-select-plus supports defaultValue, right?

No, sadly it doesn't :(

> I try to avoid setting state inside the render path if at all possible

Same. React docs say it's better to do it in the render path than in a useEffect though:

> [Storing information from previous renders](https://beta.reactjs.org/reference/react/useState#storing-information-from-previous-renders) like this can be hard to understand, but it’s better than updating the same state in an Effect. In the above example, setSelection is called directly during a render. React will re-render the List immediately after it exits with a return statement. By that point, React hasn’t rendered the List children or updated the DOM yet, so this lets the List children skip rendering the stale selection value.

infiniteluke: Haha yet another reason to move away from tact select plus. Thx for explaining

infiniteluke (approved):

PR #1130, opened 2023-01-25
Convert tributejsto dynamic import

steventhan:


petermiller310 (approved):

JeremyRH (approved):
I don't like it but I can't think of a better way to make this work. Some consumers of react-gears might have trouble with dynamic imports but webpack handles them pretty well, so who knows.

PR #1133, opened 2023-01-30
Allow other properties in FilterType

infiniteluke:
I have a use case for passing some other properties that `onRemove` need in order. the `label` and `value` are not enough to know what to do in my `onRemove`

### DIFF
--- a/src/components/FilterList/FilterList.d.ts
+++ b/src/components/FilterList/FilterList.d.ts
@@ -4,6 +4,7 @@ interface FilterType {
   label: string;
   value: string;
   removable?: boolean;
+  [index: string]: any;
 }
 
 interface FilterListProps {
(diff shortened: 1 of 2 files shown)


src/components/FilterList/FilterList.d.ts
```diff
@@ -4,6 +4,7 @@ interface FilterType {
   label: string;
   value: string;
   removable?: boolean;
+  [index: string]: any;
```
steventhan, line 7: `{ label: 'hello', value: 'world', myVal: 'value' }` should already be compatible without this line due to TS's structural type system, is it not the case for you?

infiniteluke: Thank you! I had a different issue!

PR #1137, opened 2023-02-24
270 react-gears - add icon style to include cc brands

fabio-aiello-appfolio:
### Add icon style

Add 'brands' to iconStyle props and storybook to correctly display FontAwesome 6 credit card icons.


emcpete (approved):
Looks good

PR #1139, opened 2023-03-13
fix: onSelect should not be called when the List selected property is updated

frmalinowski:
**Overview**
Fixed a bug where the `onSelect` property for the `List` component (and components that inherit from `List` such as `SortableList`) is called when the `selected` attribute changes.

The parent component of `List` triggers the change of the `selected` property so it does not need to be aware of the selection change.

Only when a checkbox, switch or radio button in the `List` is selected / unselected should the `onSelect` callback be triggered.

### DIFF
--- a/src/components/List/List.tsx
+++ b/src/components/List/List.tsx
@@ -94,10 +94,6 @@ function List<T extends Item>({
   const [selectAllId] = useState(() => uniqueId('selectall-'));
   const selectAllRef = useRef<HTMLInputElement>(null);
 
-  useDeepCompareEffect(
-    () => onSelect(Array.from(selection.values())),
-    [Array.from(selection.values()), onSelect]
-  );
   useDeepCompareEffect(() => replaceSelection(selected), [selected, replaceSelection]);
 
   useDeepCompareEffect(() => {
@@ -117,16 +113,19 @@ function List<T extends Item>({
   }, [items, Array.from(selection.values())]);
 
   const handleSelection = (item: T, checked?: boolean) => {
+    let newSelection = selection;
+
     if (select === 'checkbox' || select === 'switch') {
       if (hasItem(item) && !checked) {
-        removeItem(item);
+        newSelection = removeItem(item);
       } else if (checked) {
-        addItem(item);
+        newSelection = addItem(item);
       }
     } else if (!hasItem(item)) {
       clearSelection();
-      addItem(item);
+      newSelection = addItem(item);
     }
+    onSelect(Array.from(newSelection.values()));
   };
 
   const handleSelectAll = () => {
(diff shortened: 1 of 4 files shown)


frmalinowski:
[Avoid List component in React Gears to trigger onSelect callback when the selected property has changed](https://trello.com/c/nJ7ndf2q/614-avoid-list-component-in-react-gears-to-trigger-onselect-callback-when-the-selected-property-has-changed)

src/hooks/useMap.ts
```diff
@@ -6,17 +6,20 @@ export default function useMap<T>(defaultValue: T[], keyMapper: (value: T) => an
   const add = (value: T) => {
     map.set(keyMapper(value), value);
     setMap(new Map(map));
+
+    return map;
```
JeremyRH, line 10: This should return the new map but is returning the old, mutated map.

src/components/List/List.tsx
```diff
@@ -94,10 +94,6 @@ function List<T extends Item>({
   const [selectAllId] = useState(() => uniqueId('selectall-'));
   const selectAllRef = useRef<HTMLInputElement>(null);
 
-  useDeepCompareEffect(
-    () => onSelect(Array.from(selection.values())),
-    [Array.from(selection.values()), onSelect]
-  );
```
JeremyRH, line 100: Removing this could break code relying on it. Existing usage of List and might expect `onSelect` to be called on first render, when passing in new `selected` items, and passing in a new `onSelect` callback.

src/hooks/useMap.ts
frmalinowski: That's a good point, I will modify it to return the new map.

src/components/List/List.tsx
frmalinowski: In that case, code that relies on it needs to be modified.
It does not look right to have `onSelect` called on the first render when no action was done: the parent component passes down the `selected` property so why does it need to know the properties that were just passed down?

frmalinowski: *the items that were just passed down

JeremyRH: I agree completely. It could be easy to update the code that relies on this or not. I haven't looked yet.

frmalinowski:
Thank you Jeremy for your review.

I have updated [`useMap.ts`](https://github.com/appfolio/react-gears/pull/1139/files#diff-9ebad40b365f7016ea458a78008cb8d743371f292c86b3ba95577960b7dd19d8) and the test file [`useMap.spec.ts`](https://github.com/appfolio/react-gears/pull/1139/files#diff-33fa8f8c781488c617cc53c1fcafc4fd9ffc7e0bc5c41e2312f9dd685bf3a0a3) to return the reference to the new map and not the old one.

src/components/List/List.tsx
frmalinowski: Yes it could potentially break some existing code.

frmalinowski:
Thank you Jeremy!
I updated the commit message to `fix!:` as you said.

JeremyRH (approved):

PR #1144, opened 2023-03-27
289 react-gears add aria-required to inputs

fabio-aiello-appfolio:
### Add "required" attribute to required input fields

include `aria-required="true"` in input components where entry is required


fabio-aiello-appfolio:
[Add "required" attribute to required input fields](https://trello.com/c/uus8DQw6/289-add-required-attribute-to-required-input-fields)

PR #1154, opened 2023-04-27
Clean up docs and files

JeremyRH:


aaronmars (approved):

.github/workflows/commitlint.yml
```diff
@@ -1,11 +0,0 @@
-name: Lint Commit Messages
```
davidacevedo, line 1: What is this being replaced with? Others that don't usually use conventional commits end up using it because this fails for them.

davidacevedo: Specifically, what is the github settings?

JeremyRH: [GitHub rulesets](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets).

It does not function exactly the same. The user only sees the message when clicking the merge button on their PR.
The error message is customizable. It currently says to use conventional commits and provides a link to the conventional commits website.

PR #1158, opened 2023-05-05
CollapsableText: Allow for any children, move show more button to bottom

JeremyRH:
Before:
Screenshot 2023-05-04 at 5 55 22 PM

After:
Screenshot 2023-05-06 at 7 42 05 PM

This is a soft-breaking change. `maxLength` prop is not respected. It's now `maxLines`.

### DIFF
--- a/src/components/CollapsableText/CollapsableText.spec.js
+++ b/src/components/CollapsableText/CollapsableText.spec.js
@@ -1,72 +1,74 @@
-import assert from 'assert';
-import { mount } from 'enzyme';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import React from 'react';
-import Button from '../Button/Button';
 import CollapsableText from './CollapsableText';
 
+const mockElement = {
+  clientHeight: 1,
+  scrollHeight: 2,
+};
+
+jest.mock('../../hooks/useIntervalRef', () => {
+  return {
+    useIntervalRef: jest.fn((cb) => () => cb(mockElement)),
+  };
+});
+
 describe('<CollapsableText />', () => {
-  it('should show whole string without button if string length is smaller than maxLength', () => {
-    const component = mount(<CollapsableText>Hello World</CollapsableText>);
-    assert.equal(component.text(), 'Hello World');
+  beforeEach(() => {
+    mockElement.scrollHeight = 2;
   });
 
-  it('should shorten the string and show button if string length is greater than maxLength', () => {
-    const component = mount(<CollapsableText maxLength={5}>Hello World</CollapsableText>);
-    assert.equal(component.text(), 'Hello… Show More');
+  it('should show the whole string without button if line length is less than maxLines', () => {
+    mockElement.scrollHeight = 1;
+    render(<CollapsableText maxLines={1}>Hello World</CollapsableText>);
+    expect(screen.getByText('Hello World')).toBeInTheDocument();
+    expect(screen.queryByText('Show More')).toBeNull();
   });
 
-  it('should show whole string and show button if string length is greater than maxLength and default is to show all', () => {
-    const component = mount(
-      <CollapsableText maxLength={5} collapsed={false}>
-        Hello World
+  it('should limit the container style.maxHeight and show a "Show More" button', () => {
+    render(
+      <CollapsableText maxLines={1}>
+        Hello World <br /> Wow
       </CollapsableText>
     );
-    assert.equal(component.text(), 'Hello World Show Less');
+    expect(screen.getByText(/Hello World/).style.maxHeight).toBeTruthy();
+    expect(screen.getByText('Show More')).toBeInTheDocument();
   });
 
-  it('toggle show whole or part of string after clicking button', () => {
-    const component = mount(<CollapsableText maxLength={5}>Hello World</CollapsableText>);
-    assert.equal(component.text(), 'Hello… Show More');
-
-    let button = component.find(Button);
-    button.simulate('click');
-    assert.equal(component.text(), 'Hello World Show Less');
-
-    button = component.find(Button);
-    button.simulate('click');
-    assert.equal(component.text(), 'Hello… Show More');
-  });
-
-  it('should toggle after prop change', () => {
-    const component = mount(<CollapsableText maxLength={5}>Hello World</CollapsableText>);
-    assert.equal(component.text(), 'Hello… Show More');
-
-    component.setProps({ collapsed: false });
-    component.update();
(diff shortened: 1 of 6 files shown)


src/components/CollapsableText/CollapsableText.spec.js
```diff
@@ -1,72 +1,74 @@
...
-        Hello World
+  it('should limit the container style.maxHeight and show a "Show More" button', () => {
+    render(
+      <CollapsableText maxLines={1}>
+        Hello World <br /> Wow
       </CollapsableText>
     );
-    assert.equal(component.text(), 'Hello World Show Less');
+    expect(screen.getByText(/Hello World/).style.maxHeight).toBeTruthy();
+    screen.getByText('Show More');
```
chrism6475, line 36: [nit] do you want an assertion here for the 'Show More' button?
`expect(screen.queryByText('Show More')).toBeNull()`

src/components/CollapsableText/CollapsableText.tsx
```diff
@@ -1,59 +1,73 @@
...
-  }
-  if (collapsed) {
-    return (
-      <span>
-        {children.substring(0, maxLength).trim()}&hellip;{' '}
-        <Toggle onClick={() => toggle()}>{moreLabel}</Toggle>
-      </span>
-    );
+  // If the collapsed prop changes, update the state.
+  // This is bad practice, we should instead have a defaultCollapsed prop and the collapsed prop
```
chrism6475, line 44: I agree. Do you want to add a TODO to address this down the line or do you intend to refactor collapsed to be controlled via prop in this PR?

chrism6475:
Whoa you're speedy with PR's - thanks for putting this all together. I tested locally and this all looks good with NewKids' use case. I was wondering if we want to left align the "Show More" button with the text: https://user-images.githubusercontent.com/15149443/236867790-2b1acad1-10d5-4a89-a36c-957066da5d44.mov

src/components/CollapsableText/CollapsableText.spec.js
JeremyRH: `getByText` will throw if the element can't be found. I've gotten used to this style but agree it is more clear to use an assertion.

src/components/CollapsableText/CollapsableText.tsx
JeremyRH: I want to limit the impact of this PR. I'll add a TODO.

src/hooks/useOnResizeRef.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (el && !elementIsObserved(el, cb) && typeof ResizeObserver !== 'undefined') {
+      addObservedElement(el, cb);
+
+      const throttledCb = throttle(cb, 100);
+      const resizeObserver = new ResizeObserver(() => throttledCb(el));
+
+      resizeObserver.observe(el);
+
+      // ResizeObserver doesn't catch resizes due to font size changes.
+      // This is a hack to catch those resizes.
```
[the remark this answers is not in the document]
JeremyRH: `font-size` changes outside of the first 10 seconds of the element being observed are not caught. This was a frustrating issue and I almost scrapped using `ResizeObserver` in favor of only using an interval (at a faster interval than 1 second).
The callback is throttled to prevent it being called more than once in 100ms. It shouldn't matter how many times the callback is called, resizes can happen any number of times and at any time.

JeremyRH:
> Whoa you're speedy with PR's - thanks for putting this all together. I tested locally and this all looks good with NewKids' use case. I was wondering if we want to left align the "Show More" button with the text: https://user-images.githubusercontent.com/15149443/236867790-2b1acad1-10d5-4a89-a36c-957066da5d44.mov

@ashleypresman requested the button be left aligned and made larger.

src/components/CollapsableText/CollapsableText.spec.js
chrism6475: Gotcha, yea this works! 

I only mention because I recalled from kent dodd's tutorial that he [suggests](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-get-variants-as-assertions) adding an assertion for readability like you said. 

I've also personally only used `getBy*` to insure that an element critical to the functionality of the test is present (e.g. get 'Show More' button to click and verify the full text is displayed) and `queryBy*` for assertions, but tbh I have to familiarize myself with how y'all use react testing library.

src/components/CollapsableText/CollapsableText.tsx
chrism6475: Makes sense!

src/hooks/useOnResizeRef.ts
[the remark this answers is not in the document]
chrism6475: Darn, that's such an annoying limitation of the observer. I think your interval is fine. I imagine it's unlikely for a user to change the font-size once they're settled on the page after that window anyways.

Ah thanks for explaining! I missed the throttle, thanks!

chrism6475:
> > Whoa you're speedy with PR's - thanks for putting this all together. I tested locally and this all looks good with NewKids' use case. I was wondering if we want to left align the "Show More" button with the text: https://user-images.githubusercontent.com/15149443/236867790-2b1acad1-10d5-4a89-a36c-957066da5d44.mov
> 
> @ashleypresman requested the button be left aligned and made larger.

Oh gotcha ya, I thought there was some padding around the "Show More" button, but I think looks alright.

chrism6475 (approved):
Feel free to add someone else as a reviewer or keep open if this was just a POC, but these changes lgtm

src/hooks/useOnResizeRef.ts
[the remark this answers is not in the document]
JeremyRH: I was wrong about the `ResizeObserver`. It does catch resizes due to font changes. The issue is it does not observe changes that do not affect the elements actual size. This makes sense, can be a problem:

1. The container for `children` starts exactly at `maxLines` and fits everything.
2. We check there is no overflow (`scrollHeight` same as `clientHeight`) and hide the "Show More" button.
3. A font loads, making the text larger, which adds a new line.
4. The container does not change size, it was already at `maxLines`. Yes, its `scrollHeight` changes, but `ResizeObserver` does not observe this because this is not part of the element's visible size.
5. `ResizeObserver` callback is not called and the "Show More" button remains hidden.

This is a rare edge case but also a pretty bad one if it does happen. I decided to use an interval to continuously check the `scrollHeight` instead of `ResizeObserver`. It's set to check every 200ms. Shouldn't be too bad for performance.

PR #1162, opened 2023-05-20
Add GitHub link to stories

theovelikov:
This PR adds a link to each components source code so it can easily be accessed from the stories page

Screenshot 2023-05-20 at 12 32 03 AM

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -15,6 +15,7 @@ module.exports = {
     },
     '@storybook/addon-knobs',
     '@storybook/addon-actions',
+    'storybook-source-link',
     {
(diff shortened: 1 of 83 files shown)


JeremyRH (changes requested):
Thanks for doing this!
Updating packages might have caused type issues. Can you check those files failing to compile in CI and see if the types need to be updated?
src/components/Checkbox/CheckboxListInput.tsx
src/components/Combobox/Combobox.tsx
src/components/Form/FormRow.tsx
src/components/List/SortableList.tsx

JeremyRH (approved):

PR #1165, opened 2023-06-06
Allow InfoBox title to be an Element

kkirasich:
Dude team would like the title to be a link on the InfoBox.

### DIFF
--- a/src/components/InfoBox/InfoBox.tsx
+++ b/src/components/InfoBox/InfoBox.tsx
@@ -2,12 +2,12 @@ import classnames from 'classnames';
 import React, { ReactNode } from 'react';
 import Icon from '../Icon/Icon';
 
-interface InfoBoxProps extends Omit<React.HTMLProps<HTMLDivElement>, 'className'> {
+interface InfoBoxProps extends Omit<React.HTMLProps<HTMLDivElement>, 'className' | 'title'> {
   children?: ReactNode;
   className?: string;
   color?: string;
   icon?: string;
-  title?: string;
+  title?: string | Element;
   vertical?: boolean;
 }
 


kkirasich:
https://trello.com/c/BS4mQijs/2471-new-dashboard-item-total-disbursements

src/components/InfoBox/InfoBox.tsx
```diff
@@ -2,12 +2,12 @@ import classnames from 'classnames';
 import React, { ReactNode } from 'react';
 import Icon from '../Icon/Icon';
 
-interface InfoBoxProps extends Omit<React.HTMLProps<HTMLDivElement>, 'className'> {
+interface InfoBoxProps extends Omit<React.HTMLProps<HTMLDivElement>, 'className' | 'title'> {
```
gcpreston, line 5: What does this do?

kkirasich: So we basically get the error that 
```
Types of property 'title' are incompatible.
    Type 'string | Element | undefined' is not assignable to type 'string | undefined'.
      Type 'Element' is not assignable to type 'string'.
```
because `title` is a property in the extended interface of `Omit<HTMLProps<HTMLDivElement>`. So adding this allows title's types to be overridden below in this block.

JeremyRH (approved):

src/components/InfoBox/InfoBox.tsx
```diff
@@ -2,12 +2,12 @@ import classnames from 'classnames';
...
   children?: ReactNode;
   className?: string;
   color?: string;
   icon?: string;
-  title?: string;
+  title?: string | Element;
```
JeremyRH, line 10: Should this be `ReactNode` or `Element`?

kkirasich: Our use case is just an `Element` but curious if `ReactNode` will work for basic HTML elements as well and if that is better for future use cases?

crgb0s (approved):
🐸

src/components/InfoBox/InfoBox.tsx
JeremyRH: [`ReactNode`](https://reactnative.dev/docs/react-node) works for all things that might be `children` of a component.

kkirasich: Yeah actually I do need it

PR #1170, opened 2023-06-14
Makes passed style props additive rather than replacing in tooltipbutton

KjWhal:
Working on components for snoop-doc revealed that style props passed in would overwrite an existing style prop necessary for properly handling disabled state. 

This change makes the passed in style props complement, rather than overwrite the existing behavior


JeremyRH:
Thanks! Just one thing, your commit needs to follow conventional commits so the version gets bumped automatically:
https://github.com/appfolio/react-gears#commit-messages

JeremyRH (approved):

PR #1171, opened 2023-06-15
Align Modal Text Color with Design System

dshen94:
Removes logic to turn field text to red color when the field is required in order to align with design system. The text will now be rendered in black while the red asterisk is kept as the required indicator.


JeremyRH (approved):

JeremyRH:
Thanks!
One thing: commit messages need to be lower-case.

PR #1184, opened 2023-08-29
feat: allow labelClassName to be overridden

theovelikov:
Sometimes users want to be able to use custom spacing for their labels. This gives them that flexibility

### DIFF
--- a/src/components/Form/FormLabelGroup.tsx
+++ b/src/components/Form/FormLabelGroup.tsx
@@ -27,6 +27,7 @@ type FormLabelGroupProps = React.PropsWithChildren<{
   inputId?: string;
   label?: React.ReactNode;
   labelSize?: keyof typeof labelSizeTranslations;
+  labelClassName?: string;
   required?: boolean;
   rowClassName?: string;
   size?: string;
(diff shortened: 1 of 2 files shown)


src/components/Form/FormLabelGroup.tsx
```diff
@@ -65,12 +67,15 @@ function FormLabelGroup({
...
-  });
+  let labelClassNames = labelClassName;
+  if (!labelClassName) {
+    labelClassNames = classnames({
+      'text-sm-end pe-0': !stacked,
+      'text-danger': feedback,
+      'text-success': validFeedback,
+      'visually-hidden visually-hidden-focusable': srLabel,
+    });
+  }
```
steventhan, line 78: Nit: use coalescing operator here
```suggestion
  const labelClassNames = labelClassName ?? classnames({
    'text-sm-end pe-0': !stacked,
    'text-danger': feedback,
    'text-success': validFeedback,
    'visually-hidden visually-hidden-focusable': srLabel,
  }); 
```

steventhan (approved):

src/components/Form/FormLabelGroup.tsx
theovelikov: Ohhh nice! Didn't know about this, thanks

PR #1190, opened 2023-09-15
Add `disabled` row option to disable input on specific rows

CRiva:
Affects Uncontrolled and Sortable tables

SelectableRows

### DIFF
--- a/src/components/Table/SortableTable.js
+++ b/src/components/Table/SortableTable.js
@@ -86,6 +86,7 @@ function getSelectableCell(row, rowSelected, onSelect) {
         checked={rowSelected(row)}
         onClick={(e) => e.stopPropagation()}
         onChange={(e) => onSelect(row, e.target.checked)}
+        disabled={!!row.disabled}
       />
     </>
   );
--- a/src/components/Table/UncontrolledTable.js
+++ b/src/components/Table/UncontrolledTable.js
@@ -78,7 +78,8 @@ export default class UncontrolledTable extends React.Component {
   };
 
   get allSelected() {
-    return this.props.rows.length && this.state.selected.length === this.props.rows.length;
+    const selectableRows = this.props.rows.filter((row) => !row.disabled);
+    return this.props.rows.length && this.state.selected.length === selectableRows.length;
   }
 
   selected(value) {
@@ -97,7 +98,8 @@ export default class UncontrolledTable extends React.Component {
   };
 
   toggleAll = () => {
-    const newSelection = this.allSelected ? [] : this.props.rows;
+    const selectableRows = this.props.rows.filter((row) => !row.disabled);
+    const newSelection = this.allSelected ? [] : selectableRows;
 
     this.setState({ selected: newSelection }, () => {
       this.props.onSelect(newSelection);
--- a/src/components/Table/SortableTable.spec.js
+++ b/src/components/Table/SortableTable.spec.js
@@ -377,6 +377,24 @@ describe('<SortableTable />', () => {
       assert.equal(wrapper.find('input').length, 4, 'select checkbox missing');
     });
 
+    it('should disable select checkbox when specified', () => {
+      const rows = [
+        { name: 'Alpha', disabled: false },
+        { name: 'Bravo', disabled: true },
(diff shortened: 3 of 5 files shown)


src/components/Table/SortableTable.js
```diff
@@ -74,6 +74,7 @@ function defaultRenderRow(
 
 function getSelectableCell(row, rowSelected, onSelect) {
   const selectRowId = uniqueId('select-row-');
+  const rowSelectable = Object.hasOwn(row, 'selectable') ? row.selectable : true;
```
JeremyRH, line 77: Does TS not complain about `Object.hasOwn`? Every time I use it in property, it complains because it's a newish API and the built in TS types don't know about it.

CRiva: Well this is a JS file. So no, but it complained when I used `row.hasOwnProperty('selectable')` so I figured this was the more appropriate approach. Is there another solution to this?

JeremyRH: `Object.hasOwn` is the best solution but requires the built-in lib.d.ts to be es2022 or greater and doesn't have type guards yet.

You could do `'selectable' in row. That searches the object's prototype chain so it's not the best, but works fine in this case.

You could also change `row.selectable` to `row.disabled` right? That would get rid of this code.

JeremyRH (approved):

src/components/Table/SortableTable.js
CRiva: `selectable` is just an attribute on the `row` object. It wouldn't matter whether I used `selectable` or `disabled` 

So I think we'd still need the logic here to ensure that either key existed on the object. 

Because if the key didn't exist, it would default to `undefined` a falsey value, therefore every row would be disabled. 

Is the `Object.hasOwn` vs the `'selectable' in row` a blocker? Would the es2022 ever become an issue going forward? 

If not I'll merge as is.

JeremyRH: > Because if the key didn't exist, it would default to undefined a falsey value

`disabled` is the reverse of `selectable` so you shouldn't need logic to detect if the key is present for `disabled`. `disabled={undefined}` should behave the same as `disabled={false}`.

> Is the Object.hasOwn vs the 'selectable' in row a blocker?

I don't think it's a blocker but it could be if an app was not using lib.d.ts es2022 or greater and did not [`skipLibCheck`](https://www.typescriptlang.org/tsconfig#skipLibCheck).

CRiva: ah, i see what you're saying now. Okay I'll see about changing that 👍

CRiva: Changed it to `disabled` from `selectable` 👍

src/components/Table/SortableTable.js
```diff
@@ -74,6 +74,7 @@ function defaultRenderRow(
 
 function getSelectableCell(row, rowSelected, onSelect) {
   const selectRowId = uniqueId('select-row-');
+  const disabled = Object.hasOwn(row, 'disabled') ? row.disabled : false;
```
JeremyRH, line 77: Does using `Object.hasOwn` fix an edge case? I can't think of how this would be different than just passing `row.disabled` to `<input>`.

src/components/Table/UncontrolledTable.js
```diff
@@ -78,7 +78,10 @@ export default class UncontrolledTable extends React.Component {
   };
 
   get allSelected() {
-    return this.props.rows.length && this.state.selected.length === this.props.rows.length;
+    const selectableRows = this.props.rows.filter((row) =>
+      Object.hasOwn(row, 'disabled') ? !row.disabled : true
```
JeremyRH, line 82: Same here, if `row.disabled` is not set, it is `undefined`. `!undefined` is `true` so the`Object.hasOwn` seems like is isn't needed.

JeremyRH (approved):

src/components/Table/SortableTable.js
CRiva: good points. I think there was something with `selectable` that it made sense here, but you're right with the disabled. 

I'll update that then merge :)

PR #1195, opened 2023-10-05
Export combobox props

andrewdt23:
This commit exports ComboboxProps so that it can be used outside of react-gears. This allows clients of react-gears to interface with the prop types for this component.


andrewdt23:
Closing this since we found a resolution without having to add this export: https://appfolio.slack.com/archives/C03EA63K80K/p1696623529694499

PR #1207, opened 2023-10-25
Update storybook CSS to new coastline URL

aaronmars:


j-ochs (approved):
Thanks for the quick fix!

PR #1213, opened 2023-12-07
DEBT-346 rg Add data testid to components

fabio-aiello-appfolio:
### Add data-testid to components

* Select components to be modified as per Jira issue


emcpete (approved):
Looks goo

PR #1214, opened 2023-12-12
fix: SortableTable colors correct after rowExpanded

khalludi:
WIP


khalludi:
No longer pursuing this change.

PR #1233, opened 2024-04-22
feat(APT-1610): add portalEl prop to Combobox

chasingmaxwell:
This utilizes [createPortal](https://react.dev/reference/react-dom/createPortal) to allow space/position constrained implementations of `Combobox` to render the dropdown menu in another component so the positioning of the menu is not dependent on the parent element.

In BulkUploader we need this functionality to prevent menus from being hidden behind table headers:

before: (because the dropdown menu could not be displayed outside the bounds of the table, popperjs positioned the menu on top, but it was hidden under the sticky header and couldn't be positioned above it because of the positioning of virtualized rows)
image

after: (using the `portalEl` prop we can render the dropdown menu as a child of `document.body` which allows popper to position it below and allow it to extend outside the boundary of the table regardless of the positioning of its parent row)
image

### DIFF
--- a/src/components/Combobox/Combobox.spec.js
+++ b/src/components/Combobox/Combobox.spec.js
@@ -1,5 +1,5 @@
 import assert from 'assert';
-import { render, fireEvent, cleanup } from '@testing-library/react';
+import { render, fireEvent, cleanup, screen, within } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import React from 'react';
 import sinon from 'sinon';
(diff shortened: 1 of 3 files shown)


bryan-ash:
It would be super helpful if you could add a story to the storybook, and drop a screenshot in the PR description.

aaronmars (approved):

chasingmaxwell:
@bryan-ash I've added a story and screenshots to the PR description. Thanks!

bryan-ash (approved):
☀️

PR #1237, opened 2024-05-09
feat(CollapsableText): add autoAlignToggleButton prop to CollapsableText

mdalpozzo:


JeremyRH (approved):

JeremyRH (approved):
Looks good. Let's merge it. A follow-up change can default `alignToggleButton` to `auto`. That way, if it breaks things, reverting it is no problem.

PR #1248, opened 2024-08-08
ScrollContainer: Replace broken sample image in docs

roomonthebroom:
Replaces the previous broken image in the ScrollContainer docs. Uses the same image that's used in the Save Scroll Position example so user can see the difference between the setups.

Before:
image

After:
image


davidacevedo (approved):
GG EZ 🔥

PR #1255, opened 2024-08-29
tvk fix summary box types

tkissingappfolio:
- **Fix missing yarn entry in .tool-versions**
- **Fix label and value types on SummaryBoxItem**


src/components/SummaryBox/SummaryBoxItem.tsx
```diff
@@ -6,8 +6,8 @@ import CardBody from '../Card/CardBody';
 
 interface SummaryBoxItemProps extends Omit<CardProps, 'color' | 'outline' | 'className'> {
   className?: string;
-  label?: string;
-  value?: React.ReactNode;
+  label?: string | React.ReactNode;
+  value?: string | React.ReactNode;
```
JeremyRH, line 10: Isn't `ReactNode` already a type union with `string`? Maybe different versions of `@types/react` are different.

JeremyRH (approved):

src/components/SummaryBox/SummaryBoxItem.tsx
tkissingappfolio: You are right, let me remove `string`

PR #1262, opened 2025-01-30
ahres-2168 Remove invalid options for combobox component

sidiinia:
Fix: Ensure selected options array does not contain undefined values

### DIFF
--- a/src/components/Combobox/Combobox.tsx
+++ b/src/components/Combobox/Combobox.tsx
@@ -95,7 +95,9 @@ function Combobox<T>({
   const selected = useMemo<Option<T> | Option<T>[]>(
     () =>
       multi
-        ? (value || []).map((v: T) => options.find((option) => equal(option.value, v)))
+        ? (value || [])
+            .map((v: T) => options.find((option) => equal(option.value, v)))
+            .filter((option: Option<T>) => option !== undefined)
         : options.find((option) => equal(option.value, value)),
     [value, options, multi]
   );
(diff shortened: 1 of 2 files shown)


JeremyRH (approved):

src/components/Combobox/Combobox.tsx
```diff
@@ -95,7 +95,9 @@ function Combobox<T>({
   const selected = useMemo<Option<T> | Option<T>[]>(
     () =>
       multi
-        ? (value || []).map((v: T) => options.find((option) => equal(option.value, v)))
+        ? (value || [])
+            .map((v: T) => options.find((option) => equal(option.value, v)))
```
mdalpozzo, line 99: you could write this a bit more efficiently with `reduce` if you wanted, but also fine as is

```
.reduce<Option<T>[]>((acc, v) => {
      const found = options.find((option) => equal(option.value, v));
      if (found) acc.push(found);
      return acc;
    }, [])
```

ztmeyer (approved):

DistractedPrimate:
Open question (non blocking): Is there a way that we can somehow log when the combobox is receiving invalid parameters? Would be nice to know when we are passing bad params in a way that doesn't crash pages.

JeremyRH:
> Open question (non blocking): Is there a way that we can somehow log when the combobox is receiving invalid parameters? Would be nice to know when we are passing bad params in a way that doesn't crash pages.

@DistractedPrimate TypeScript should handle this. Logs sometimes get ignored.

JeremyRH:
@sidiinia Your commit messages must follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). E.g. `Fix: Ensure selected options array does not contain undefined values`

PR #1264, opened 2025-02-03
[FEE-927] add offset on Tooltip popover content

Brian-Holland:
Our current Tooltip implementation does not have much, if any, space between the tooltip arrow and the content it's pointing to. This fix adds an offset based on the placement direction.

Before:
Screenshot 2025-02-03 at 5 13 33 PM
Screenshot 2025-02-03 at 5 09 46 PM
Screenshot 2025-02-03 at 5 13 44 PM
Screenshot 2025-02-03 at 5 13 51 PM

After:
Screenshot 2025-02-03 at 5 12 37 PM
Screenshot 2025-02-03 at 5 14 45 PM
Screenshot 2025-02-03 at 5 12 45 PM
Screenshot 2025-02-03 at 5 14 56 PM

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -9,6 +9,7 @@ const Tooltip: FC<TooltipProps> = ({
   isOpen = defaultProps.isOpen,
   fade = defaultProps.fade,
   target,
+  placement = 'top',
   ...props
 }) => {
   const [open, setOpen] = useState(isOpen);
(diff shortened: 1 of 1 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -17,8 +18,26 @@ const Tooltip: FC<TooltipProps> = ({
     setOpen(!open);
   }, [open]);
 
+  //adds offset for the popper placement base on placement prop
+  const offsets: Record<string, [number, number]> = {
+    top: [0, 5],
+    bottom: [0, 5],
+    left: [0, 8],
+    right: [0, 8],
+  };
```
Brian-Holland, line 27: Offsets are uneven here to account for the vertical space around text that isn't there horizontally.

JeremyRH (approved):

PR #1272, opened 2025-06-05
Render note.from as react component instead of a string

fldm713:
This pull request includes a minor adjustment to the `NoteHeader` component in `src/components/Note/NoteHeader.tsx`. The change simplifies the JSX structure for displaying the "by" text in the note header.

* [`src/components/Note/NoteHeader.tsx`](diffhunk://#diff-18b3e46bdb2438ccd781986d7d047001e220f555ab9cb36653009a9fd0c39bc3L59-R59): Updated the `from` rendering logic to remove unnecessary template literals and simplify the JSX structure for the "by" text.


src/components/Note/NoteHeader.tsx
```diff
@@ -20,14 +20,15 @@ type NoteHeaderProps = {
   showTimezone?: boolean;
   onDelete?: (note: Omit<Note, 'text'>) => void;
   onEdit?: (note: Omit<Note, 'text'>) => void;
+  color?: string;
```
aaronmars, line 23: I noticed that color defaults to "info" in 2 other usages, and is optional here. I think it would be better to put the "info" default here so that the ```` `bg-${color}-subtle` ```` has no problems.

fldm713: I made the change

aaronmars (approved):

PR #1310, opened 2026-03-16
OWNER-221: fix DateInput not clearing when controlled value becomes undefined

scottpchow23:
## Summary
- Fixes a bug where `DateInput` did not clear its displayed value when the controlled `value` prop changed from a date string to `undefined` or `''`. This specifically happened when using `reset` on a form using `react-hook-forms` with a controlled `DateInput`.
- Adds reset logic in `componentDidUpdate` to clear both internal state and the DOM input element when `value` transitions from truthy to falsy
- Adds tests for both `undefined` and empty string reset scenarios

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/components/Input/DateInput.js
+++ b/src/components/Input/DateInput.js
@@ -248,6 +248,13 @@ export default class DateInput extends React.Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
+    if (prevProps.value && !this.props.value) {
+      this.setState({ value: '' });
+      if (this.inputEl) {
+        this.inputEl.value = '';
+        this.inputEl.setAttribute('value', '');
+      }
+    }
     this.setInputValue();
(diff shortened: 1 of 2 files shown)


aaronmars (approved):

JeremyRH (changes requested):
This can cause existing uses to break. Even if you check for a previous truthy prop value, it's possible some teams rely on the behavior of it keeping an old value around.

I agree that the current behavior is not great. If you are willing to do the work to make sure this doesn't break anything in Property, and help fix cases where it does break something, we can merge this.

scottpchow23:
Hmm fair point about this potentially breaking other things. I'm willing to help with any potential fixes around Property but I can't commit to doing an earnest regression of all date inputs in property prior to this change. As such, I'm going to leave this alone until I can actually follow through on this change in its entirety.

PR #1311, opened 2026-04-02
[OWNER-226] Make FeatureBanner subtitle optional

appf-mike:
## Summary
- Make `subtitle` prop optional on `FeatureBanner`
- When subtitle is not provided, the `<p>` element is not rendered and the title/children are vertically centered
- Supports condensed banner views such as on mobile

## Test plan
- [x] Added tests for rendering without subtitle (title and children still render)
- [x] Verify Storybook `WithoutSubtitle` story renders correctly (title + button, vertically centered)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/components/FeatureBanner/FeatureBanner.spec.tsx
+++ b/src/components/FeatureBanner/FeatureBanner.spec.tsx
@@ -39,6 +39,25 @@ describe('<FeatureBanner />', () => {
     expect(queryByText('FeatureBannerChildElement')).not.toBeNull();
   });
 
+  describe('without subtitle', () => {
+    it('does not render subtitle', () => {
+      const { queryByText } = render(<FeatureBanner title="FeatureBannerTitle" />);
+
+      expect(queryByText('FeatureBannerTitle')).not.toBeNull();
+    });
+
+    it('renders title and children', () => {
+      const { queryByText } = render(
+        <FeatureBanner title="FeatureBannerTitle">
+          <span>FeatureBannerChildElement</span>
+        </FeatureBanner>
+      );
+
+      expect(queryByText('FeatureBannerTitle')).not.toBeNull();
+      expect(queryByText('FeatureBannerChildElement')).not.toBeNull();
+    });
+  });
+
   describe('when dismissable', () => {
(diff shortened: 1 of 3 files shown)


aaronmars:
Can't this be achieved with CSS utility classes from Bootstrap? The breakpoints are set up for mobile screen widths.

appf-mike:
@aaronmars sorry I was still deciding how I wanted to do this so I waited to respond.
The actual component I want to update is the FeedbackBanner in react-apm, which I want to create a condensed version of.  I think the better change to make in this repo is to just allow a display of the component where subtitle is optional since this is less breaking for existing consumers.

Likewise in react-apm I plan to add a net new prop since it better encapsulates display permutations of the components which can be controlled by the consumer of the component, who can decide to use the condensed view whenever they want (In my case it will be for mobile display).  

I'm not sure I'm familiar with the utility classes you are prescribing -- maybe you could describe/give an example so I can evaluate that solution?

aaronmars:
Making subtitle optional here is fine, that's a reasonable API change for react-gears.

What I was getting at is for the mobile behavior in react-apm. Instead of a new prop to toggle a "condensed" view, you can use Bootstrap's responsive display classes to hide the subtitle on small screens:

```jsx
<FeatureBanner
  title="..."
  subtitle={<span className="d-none d-md-block">{subtitleText}</span>}
>
  ...
</FeatureBanner>
```

`d-none d-md-block` hides the element on mobile and shows it at the `md` breakpoint and up. That way CSS handles the responsive behavior instead of the consumer needing to manage it with a prop. Bootstrap has these for all the breakpoints: https://getbootstrap.com/docs/5.3/utilities/display/#hiding-elements

One thing to note: the current FeatureBanner always renders the `<p>` wrapper around subtitle, so even with hidden content it'll take up a bit of space from line-height. Your change here to skip rendering the `<p>` when subtitle is absent would fix that. So this PR + responsive classes in react-apm is probably the right combo.

I wasn't sure if this feedback belonged in the react-apm PR or here, but it had to go somewhere!

PR #1334, opened 2026-09-17
fix: bump Storybook 9 -> 10 to drop vulnerable @vitest/mocker

davidacevedo:
Dependabot sweep — Medium tier. One package, one commit: bump plus the required code change.

## Alert closed

| Package | Old | New | Alert | GHSA | Severity |
|---|---|---|---|---|---|
| @vitest/mocker | 3.2.4 | (removed) | [#199](https://github.com/appfolio/react-gears/security/dependabot/199) | [GHSA-82fw-gwwq-j7x9](https://github.com/advisories/GHSA-82fw-gwwq-j7x9) | medium |

## Why this needed a major parent bump

`@vitest/mocker` is not a direct dependency of react-gears; it is pulled in by `storybook`'s own `package.json`, which pins it to the **exact** version `3.2.4` (no range). `storybook@9.1.20` — this repo's declared range is `^9.1.20` — is already the newest published 9.x release, so there is no in-range parent bump that reaches the patched `4.1.11`. Storybook removed the `@vitest/mocker` dependency entirely starting at `10.6.0`. Crossing the major to `10.6.0` is the only route that closes this alert without adding an override.

## What changed

- `storybook` `^9.1.20` → `^10.6.0`
- `@storybook/addon-a11y` `^9.1.20` → `^10.6.0`
- `@storybook/addon-links` `^9.1.20` → `^10.6.0`
- `@storybook/react` `^9.1.20` → `^10.6.0`
- `@storybook/react-webpack5` `^9.1.20` → `^10.6.0`
- `@chromatic-com/storybook` `^4.1.3` → `^5.3.1` (the 4.x line does not support Storybook 10; 5.3.1's peer range is `^10.1.0`+)
- `@storybook/addon-webpack5-compiler-babel` unchanged at `^4.0.1` — it is versioned independently of Storybook's core major.

`yarn why "@vitest/mocker"` now returns nothing; the package is gone from the resolved tree.

### Code change (per the [Storybook 9→10 migration guide](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-9x-to-1000))

`.storybook/main.js` converted from CommonJS to ESM (`module.exports = {...}` → `export default {...}`) — Storybook 10 loads `.storybook/main.*` strictly as ESM. This is the only breaking change in that guide applicable to this repo:
- No local addons referenced by relative path (n/a).
- No `core.builder` override set (n/a).
- `main.js` is JavaScript, not TypeScript, so the `tsconfig.json` `moduleResolution` requirement explicitly does not apply ("JavaScript projects require no changes").
- Node 20.19+/22.12+ requirement already satisfied (this repo runs Node 24.13.1 per `.tool-versions`).
- `.storybook/preview.js` and `.storybook/manager.js` were already written in ESM syntax (Storybook bundles them through webpack regardless of the root file's module format) and needed no change.
- The removed `dev-only`/`docs-only`/`test-only` built-in tags are not used anywhere in this repo's stories.

## Testing plan

No call sites for `@vitest/mocker` anywhere in `src/` — it is Storybook/vitest internal tooling, never imported by app code. **This repo's CI (`test.yml`) runs `yarn install --immutable`, `yarn lint`, `yarn format`, and `yarn test` (jest) — it never builds or runs Storybook**, so this PR's checks cannot verify the Storybook 10 migration end to end. The Storybook docs site is only built on push to `master` via `github-pages.yml`, which is outside this PR's checks. **Recommend a human spot-check `yarn docs` (or `yarn start`) locally, or watch the first post-merge `github-pages` run, before treating the docs deploy as verified.**

## Pins retired

None — no `resolutions` entry pins `@vitest/mocker` or `storybook`.

## What this PR does not close

Nothing else; the other 9 open alerts are handled on the Easy PR (#1333).

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -12,16 +12,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@ampproject/remapping@npm:^2.2.0":
-  version: 2.3.0
-  resolution: "@ampproject/remapping@npm:2.3.0"
-  dependencies:
-    "@jridgewell/gen-mapping": "npm:^0.3.5"
-    "@jridgewell/trace-mapping": "npm:^0.3.24"
-  checksum: 10/f3451525379c68a73eb0a1e65247fbf28c0cccd126d93af21c75fceff77773d43c0d4a2d51978fb131aff25b5f2cb41a9fe48cc296e61ae65e679c4f6918b0ab
-  languageName: node
-  linkType: hard
-
 "@appfolio/eslint-config-appfolio-base@npm:^2.0.1":
   version: 2.0.1
   resolution: "@appfolio/eslint-config-appfolio-base@npm:2.0.1"
@@ -173,12 +163,12 @@ __metadata:
   linkType: hard
 
 "@babel/cli@npm:^7.24.7":
-  version: 7.24.7
-  resolution: "@babel/cli@npm:7.24.7"
+  version: 7.29.7
+  resolution: "@babel/cli@npm:7.29.7"
   dependencies:
-    "@jridgewell/trace-mapping": "npm:^0.3.25"
+    "@jridgewell/trace-mapping": "npm:^0.3.28"
     "@nicolo-ribaudo/chokidar-2": "npm:2.1.8-no-fsevents.3"
-    chokidar: "npm:^3.4.0"
+    chokidar: "npm:^3.6.0"
     commander: "npm:^6.2.0"
     convert-source-map: "npm:^2.0.0"
     fs-readdir-recursive: "npm:^1.1.0"
@@ -195,7 +185,7 @@ __metadata:
   bin:
     babel: ./bin/babel.js
     babel-external-helpers: ./bin/babel-external-helpers.js
-  checksum: 10/021bf34d7347a24a84d63c282f25d16c14da566c58326e291983877fc62de7e292267e336a753d4d50aacd16b95879aaa6b3f560407fe002969a14e68832980d
+  checksum: 10/403b9b9bb50881eceff0367949d6642ed7fb829a6e8c17a8a36dc7c5bdcecac4a4decb5dab1029624967a212f56709d513cecf2db0f8bf9e06f9370d3758a7b6
   languageName: node
   linkType: hard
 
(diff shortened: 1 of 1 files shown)


aaronmars (approved):

davidacevedo:
Closing — reclassifying alert #199 (`@vitest/mocker` 3.2.4, [GHSA-82fw-gwwq-j7x9](https://github.com/advisories/GHSA-82fw-gwwq-j7x9), medium) as **Hard** after two failed CI attempts at the Storybook 9→10 bump this branch carried:

1. **Attempt 1** (bump + `.storybook/main.js` CommonJS→ESM): `yarn lint` failed `tsc --noEmit` with `TS2307: Cannot find module 'storybook/actions'` / `'@storybook/react'` in 5 `.stories.tsx` files. Root cause: Storybook 10 dropped top-level `types`/`typesVersions` fields and now resolves types only through `package.json` `exports` "types" conditions (confirmed in `node_modules`: `storybook`'s own `package.json` exports `./actions` with a `types` condition; `@storybook/react` has no top-level `types` field at all). This repo's `tsconfig.json` used the legacy `moduleResolution: "Node"`, which can't see conditional exports.
2. **Attempt 2** (+ `tsconfig.json` `moduleResolution: "Node"` → `"Bundler"`, per the Storybook migration guide): `tsc` itself failed to start — `TS6046: Argument for '--moduleResolution' option must be: 'node', 'classic', 'node16', 'nodenext'`. This repo pins `"typescript": "^4.9.5"`; the `"bundler"` value was only added in TypeScript 5.0. The alternative values that *are* available pre-5.0 (`"node16"`/`"nodenext"`) require pairing `"module"` to the same setting (TS5095), which is a materially larger, riskier change (affects how every import/export in `src/` compiles) — outside what can be named as a single scoped edit within the two-attempt budget, and effectively means the real fix is bumping TypeScript itself first, a separate major-line change.

Reverted this branch back to `origin/master` (no diff). Full writeup, including the parent-chain reasoning for why no in-range fix exists, is in `reports/react-gears-2026-09-17.md` in the dependabot-updates workspace.

(Also observed on both CI attempts: this repo's `prerelease` job installs through a JFrog npm proxy that lags the public registry by hours-to-days for brand-new versions — `baseline-browser-mapping@2.11.25` and `chromatic@18.9.4` both hit "Registry failed to return reference" here. Unrelated to the Storybook compatibility problem, but worth knowing about for future bumps through this repo.)

